PR #10, opened 2019-01-16
Add dotenv dev dependency

chee:
It's required in the `updatebrowserstacklist` test

a screenshot of dotenv being used in the updatebrowserstacklist test


chee:
@CLAassistant yes i authorise the finanical times to sell my work of adding a dependency

PR #11, opened 2019-01-16
Sort the features by name before topographically sorting them 

chee:
The same list of features did not result always in the same bundle,
because the order of the bundle could change depending on the order the
features were resolved.

I have no idea how to test this.


JakeChampion:
If we have a test which has 3 polyfills (a,b,c) which don't depend on eachother, we can prove bundle a,b,c and a,c,b and b,a,c and b,c,a and c,a,b and ,c,ba are identical

PR #61, opened 2019-01-27
Add Array.prototype.flatMap

caub:
It's apparently still stage3 proposal, I thought it was already adopted, but I already did the PR :s 
It's implemented on Chrome, Firefox, ..

It looks easy to polyfill, something like `Array.prototype.flatMap = function (fn) { return [].concat(...this.map(fn)) }` but there are corner cases (reminds me of https://github.com/Financial-Times/polyfill-service/pull/1821), so it may be easier to use the official polyfill

Similarly we can add [`Object.fromEntries`](http://npm.im/object.fromentries) (very useful as well) and [`String.prototype.matchAll`](http://npm.im/string.prototype.matchall)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -352,6 +352,8 @@ jobs:
           --browserstack
       - run: npm run test-polyfills -- --features=String.prototype.trim
           --browserstack
+      - run: npm run test-polyfills -- --features=Array.prototype.flatMap
+          --browserstack
 workflows:
   test:
     jobs:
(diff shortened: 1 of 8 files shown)


JakeChampion:
Hi, the polyfill you are wanting to add would not work with the polyfill-library and the polyfill-library is not aware of commonjs (`require`, `module.exports`). I've implemented this [polyfill myself](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Array/prototype/flatMap/polyfill.js) but I did not make a good config.json.

You have a good config.json, if you can make a pull-request updating the config.json in https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Array/prototype/flatMap/config.json that would be really helpful and I will make sure it gets reviewed and merged.

PR #91, opened 2019-01-30
Add polyfill for Object.fromEntries and AddEntriesFromIterable abstract method

JakeChampion:


radeno:
Versions are outdated. Some browsers already implemented it
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries

garygreen:
Any news on this? Would be great to have polyfill for this, it's incredibly useful

PR #107, opened 2019-02-04
MDN states Symbol.iterator is supported in chrome 43, firefox 36, safari 10, opera 30

### LINKED ISSUE #49: Symbol.iterator property of Array broken on Chrome 49 with polyfill, but works fine without polyfill
We have users on Chome 49, since it is the last Chrome that works on Windows XP and Vista. On Chrome 49, reading the `Symbol.iterator` property of an `Array` gives `undefined` if the polyfill is active. For example `[][Symbol.iterator]()` gives an error. Without the polyfill it works fine.

JakeChampion:
Fixes #49


JakeChampion:
It looks like the native symbol.iterator in some browsers is not working correctly with DOMTokenList.prototype. This needs some investigating.

PR #110, opened 2019-02-06
fix semver to become a valid semver range

JakeChampion:
`9-12` is a semver pre-release of version 9.
We actually wanted this to be a range of 9,10,11,12. Which is done via `9 - 12`

### DIFF
--- a/polyfills/DOMTokenList/prototype/@@iterator/config.json
+++ b/polyfills/DOMTokenList/prototype/@@iterator/config.json
@@ -2,7 +2,7 @@
 	"browsers": {
 		"edge": "<13",
 		"edge_mob": "<13",
-		"ie": "9-12",
+		"ie": "9 - 12",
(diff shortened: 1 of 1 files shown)


JakeChampion:
It looks like the DOMTokenList polyfill fails the tests when run against IE 😔

polyfills/DOMTokenList/prototype/@@iterator/config.json
```diff
@@ -2,7 +2,7 @@
 	"browsers": {
 		"edge": "<13",
 		"edge_mob": "<13",
-		"ie": "9-12",
+		"ie": "9 - 12",
```
JakeChampion: is this correct?

JakeChampion: yes

PR #140, opened 2019-02-21
Make Set comply with specification by throwing TypeError if the given argument is not iterable

JakeChampion:
Here is the tests running without the fixed polyfill:
bs_win10_ie_11 0 1


JakeChampion:
Woo!

PR #146, opened 2019-02-22
Match/replace Symbol

### LINKED ISSUE #130: Symbol Edge versions incorrect
The versions here do not match the versions in the `config.json` files:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol

2 examples:
[`replace`](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Symbol/replace/config.json): `"edge": "*"`
[`match`](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Symbol/match/config.json): `"edge": "*"`

I am not sure what you guys use as a source for the actual versions.

A workaround for the short term would be to add feature detection to my api request.

sbinlondon:
Fixes #130

### DIFF
--- a/polyfills/Symbol/match/config.json
+++ b/polyfills/Symbol/match/config.json
@@ -4,17 +4,18 @@
 		"es2015"
 	],
 	"browsers": {
-		"edge": "*",
-		"edge_mob": "*",
+		"edge": "<15",
+		"edge_mob": "<15",
 		"ie": "*",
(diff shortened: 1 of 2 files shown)


EECOLOR:
I now have a report that lists edge 14 for the `match` property. So you can change it to `<14`

Issue #147, opened 2019-02-26
Cache fs `readFile` operations


bringking:
Feature Request

## What

Thanks for the fantastic library. We are using this in production for a fairly high-traffic application. We are noticing in our APM that we are performing quite a few `readFile` operations. When diving in, we noticed they are mostly coming from Polyfill Library. This isn't causing a huge problem per-se, but it is introducing time into our request that I think can be reduced. 

## Details

Looking at the code here - 
https://github.com/Financial-Times/polyfill-library/blob/master/lib/sources.js#L16

It seems we could replace this with a simple `require` call (which is sync, but cached). Or stick with the readFile approach and cache the result. This would obviously mean the files can't be updated after the app boots, but that seems like an edge case. Is this something you guys have thought about? I would be happy to submit a PR.

JakeChampion:
Thanks for opening this issue @bringking. I think we would want to go with a least-recently-used-cache (LRU-cache) instead of the synchronous require. I'll make a pull-request for this work now :-)

bringking:
@JakeChampion I don't see that code change in the npm release, looks like the old code 🤔

bringking:
@JakeChampion any idea when you might fix the release? Would love to get this out

PR #149, opened 2019-02-27
Use caches for getPolyfillMeta, listAliases, and listPolyfills

JakeChampion:
Resolves https://github.com/Financial-Times/polyfill-library/issues/147
Resolves https://github.com/Financial-Times/polyfill-library/issues/148

### DIFF
--- a/lib/sources.js
+++ b/lib/sources.js
@@ -5,6 +5,8 @@ const fs = require("graceful-fs");
 const denodeify = require("denodeify");
 const readFile = denodeify(fs.readFile);
 const readdir = denodeify(fs.readdir);
+const LRUCache = require('mnemonist/lru-cache');
+const polyfillMetaCache = new LRUCache(1000);
 
 const polyfillDirectory = path.join(__dirname, "../polyfills/__dist");
 /**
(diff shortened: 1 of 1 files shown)


lib/sources.js
```diff
@@ -13,31 +15,41 @@ const polyfillDirectory = path.join(__dirname, "../polyfills/__dist");
...
 
 /**
  * Get a list of all the polyfill aliases which exist within the collection of polyfill sources.
  * @returns {Promise<Array>} A promise which resolves with an object of all the polyfill aliases within the collection.
  */
-function listAliases() {
-	return readFile(path.join(polyfillDirectory, "aliases.json")).then(JSON.parse);
-}
+const listAliases = (function() {
+	const aliases = readFile(path.join(polyfillDirectory, "aliases.json")).then(JSON.parse);
```
bringking: Would be awesome to cache this one as well, it gets read from disk on every request as well

bringking:
@JakeChampion, this is great. Thank you! Any particular reason you want to avoid using require? Just curious

JakeChampion:
@bringking avoiding require is because it is a blocking call if it is the first time the file is being requied. If polyfill-library is used on a HTTP server and that blocking call happened, it would stop other requests from being handled at the same moment in time.

lib/sources.js
JakeChampion: Thanks for the review :-)

PR #150, opened 2019-02-27
add typescript types

zauberpony:
Hi,

This PR adds typings for typescript. I deduced them mainly from the description in the README.

I can't provide a test-plan/tests. I don't think there's a good way to test typescript definitions at all.

I used these definitions in our typescript project (with strict settings etc.), that's my testcase ;)

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,54 @@
+import {FeatureList, PolyfillOptions} from "polyfill-library";
+import {ReadStream} from "fs";
+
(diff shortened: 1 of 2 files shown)


JakeChampion:
Thank you for the PR @zauberpony, I don't think we will accept this PR as we already have JSDoc comments which provide types and would want to avoid having two places where types live (TypeScript and JSDoc).
If the JSDoc types are not sufficient, we will definitely accept a PR which improves them.

PR #184, opened 2019-04-04
Update Object.entries browser versions

radeno:
Based on
- https://caniuse.com/#feat=object-entries
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries


radeno:
Updated iOS Safari. Incorrect info on MDN. I will send them a PR.

https://webkit.org/blog/7477/new-web-features-in-safari-10-1/

radeno:
Too hot outside. Switched iOS and macOS versions
https://github.com/Financial-Times/polyfill-library/pull/222

PR #185, opened 2019-04-04
Update Object.values browser versions

radeno:
Based on
- https://caniuse.com/#search=object.values
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values


polyfills/Object/values/config.json
```diff
@@ -3,21 +3,21 @@
 		"es2017"
 	],
 	"browsers": {
-		"edge": "<15",
-		"edge_mob": "<15",
-		"ie": "9 - 14",
+		"edge": "< 12",
```
JakeChampion: This should be `<14` according to MDN.
image

```suggestion
		"edge": "< 14",
```

polyfills/Object/values/config.json
```diff
@@ -3,21 +3,21 @@
...
+		"edge_mob": "< 12",
```
JakeChampion:
```suggestion
		"edge_mob": "< 14",
```

polyfills/Object/values/config.json
```diff
@@ -3,21 +3,21 @@
...
+		"ie": "*",
 		"ie_mob": "*",
 		"chrome": "< 54",
-		"ios_chr": "*",
-		"safari": "*",
-		"ios_saf": "*",
+		"ios_chr": "< 54",
+		"safari": "< 10.1",
+		"ios_saf": "< 10.1",
```
JakeChampion: This should be `< 10.3` according to MDN
image
```suggestion
		"ios_saf": "< 10.3",
```

radeno:
@JakeChampion You are right. CanIUse  shows bad versions. I have pushed PR to their repo.
Official info https://docs.microsoft.com/en-us/microsoft-edge/dev-guide/whats-new/edgehtml-14
Edge mobile embed system rendering engine. So it is questionable if version 14 supports it or not.

PR #188, opened 2019-04-11
feat(TypedArray): expose TypedArray as polyfillable feature

### LINKED ISSUE #187: Expose Typed Arrays
[FEATURE]: Exposed Polyfill for TypedArrays

## What
I would like to explicity include the TypeArrays polyfill in my bundle. I've tried adding "_TypeArray" and "Unit8Array" to my features list and its not being included.

## Details
I can see that there is already a "TypeArray" polyfill available in this library and it is included with other polyfills that depend on it (like "Blob"). I can't figure out how to require the TypeArray polyfill explicitly. 

If somebody can point me in the right direction, I'm happy to open a PR with the proper changes that would need to be made.

#188 is a possible fix.

nicksrandall:
fixes #187


JakeChampion:
Hi @nicksrandall, it looks like you found a bug in the library to do with internal polyfills having external polyfill aliases and those aliases not working!

I've created a patch which addresses/fixes the bug and should now expose the typed-array polyfill when it is requested with any of it's aliases (`Uint8Array`, etc). https://github.com/Financial-Times/polyfill-library/pull/224

I'll be closing this pull-request in favour of the one which fixes the bug. Thank you for finding this :-)

PR #192, opened 2019-04-15
Add requestIdleCallback and cancelIdleCallback.

notlee:
This requestIdleCallback polyfill builds on ReactScheduler, which
calculates the browser's framerate and seperates the idle call into a
seperate tick:

"It works by scheduling a requestAnimationFrame, storing the time for the
start of the frame, then scheduling a postMessage which gets scheduled
after paint. Within the postMessage handler do as much work as possible
until time + frame rate. By separating the idle call into a separate
event tick we ensure that layout, paint and other browser work is counted
against the available time. The frame rate is dynamically adjusted."
https://github.com/facebook/react/blob/43a137d9c13064b530d95ba51138ec1607de2c99/packages/react-scheduler/src/ReactScheduler.js

**Current Todo**

- [x] Find a solution for [Safari 9 issues](https://github.com/Financial-Times/polyfill-library/pull/192#issuecomment-483620231).
- [x] Investigate why the test "schedules callbacks with an expired timeout first regardless of order" intermitently fails.
- [x] Add a `MessageChannel` Polyfill or fallback for browsers which do not support it.
- [x] Add a test to confirm callbacks continue to run whilst `requestAnimationFrame` is suspended (e.g. tab hidden, hidden iframe).
- [x] Add support for `cancelIdleCallback`


polyfills/requestIdleCallback/config.json
```diff
@@ -0,0 +1,29 @@
...
+        "ie_mob": "*",
+        "opera": "<34",
+        "op_mini": "*",
+        "op_mob": "*",
+        "safari": "*",
+        "firefox_mob": "<55",
+        "samsung_mob": "*"
+    },
+    "dependencies": [
+        "Date.now",
```
JakeChampion: We have a `performance.now` polyfill, is there a reason why we are using `Date.now` instead?

polyfills/requestIdleCallback/polyfill.js
```diff
@@ -0,0 +1,204 @@
...
+    // We start out assuming that we run at 30fps but then the heuristic
+    // tracking will adjust this value to a faster fps if we get more frequent
+    // animation frames.
+    var previousFrameTime = 33;
+    var activeFrameTime = 33;
+
+    function getNow() {
+        return performance && typeof performance.now === 'function' ?
+            performance.now() :
+            Date.now();
```
JakeChampion: We have a `performance.now` polyfill, is there a reason why we are using `Date.now` instead?

notlee:
Idle callbacks in Safari 9 are currently given an incorrect deadline by the polyfill. For some reason the time given to the animation frame callback is after the time in the message event listener retrieved with `performance.now`.

```
window.addEventListener('message', function (event) {
    console.log({message: performance.now() - start});
});
start = performance.now(); 
window.requestAnimationFrame(function(time) { 
    window.postMessage('a', '*'); 
    console.log({raf: time - start})
}); 
console.log({now: performance.now() - start});
```

Safari 9:
Screenshot 2019-04-16 at 11 05 52

Safari 12:
Screenshot 2019-04-16 at 11 05 49

polyfills/requestIdleCallback/polyfill.js
```diff
@@ -45,7 +45,8 @@
 
     var frameDeadline = 0;
 
-    var messageKey = 'polyfillIdleCallback' + Math.random().toString(36).slice(2);
+    var messageChannel = new MessageChannel();
```
JakeChampion: I don't think all the browsers we support in polyfill-library have MessageChannel, I know for certain that IE 9 does not have it.

notlee: That's true, a polyfill or fallback is on my todo list ☝️ I had thought I'd need to anyway to cater for IE6-7, which doesn't support `postMessage` at all, but I've just realised polyfill.io doesn't support them.

notlee:
[This Safari 9 bug](https://github.com/Financial-Times/polyfill-library/pull/192#issuecomment-483620231) was caused because the callback of `requestAnimationFrame` should be given a time which indicates when the callback was queued, but for some reason in Safari 9 it's greater than the current time (when calling `performance.now` within the raf callback). It appears to be `(now + (frame rate * 2))`. I've assumed this is the case and made changes accordingly but annoyingly don't see the logic and haven't been able to confirm. This is not an issue in Safari 10 or above.

polyfills/requestIdleCallback/polyfill.js
```diff
@@ -84,10 +84,16 @@
     function getDeadline(callbackObject) {
         var timeout = callbackObject.options.timeout;
         var added = callbackObject.added;
-        return {
-            timeRemaining: timeRemaining,
-            didTimeout: timeout ? added + timeout < performance.now() : false
-        };
+        var deadline = new global.IdleDeadline();
+        Object.defineProperty(deadline, 'didTimeout', {
+            get: function() {
```
JakeChampion: getters will not work in IE8 unless they are attached to a DOM node. We would need an alternative approach for IE8 (we can keep getters for other browsers).

polyfills/requestIdleCallback/polyfill.js
```diff
@@ -233,4 +239,21 @@
         return ++idleCallbackIdentifier;
     };
 
+    global.IdleDeadline = function IdleDeadline() {
+        // TODO: how to polyfill illegal constructor?
+        // throw new TypeError('Illegal constructor');
```
JakeChampion: That would work :-)

notlee:
Looks like my approach to Safari 9 wasn't quite right. I'll take another look tomorrow and then this will be ready for review.

polyfills/requestIdleCallback/polyfill.js
notlee: I was using `new` to construct instances and wasn't sure what to do about that, but got there 👍

notlee:
I've updated the implementation for Safari 9, still under the assumption that the `requestAnimationFrame` time given is for some reason `(now + (frame rate) * 2)` (unverified).

Using this assumption the results seem reasonable. E.g. at 60fps the remaining time is ~18ms, which is reduced when busying the main thread with a `while` loop.
Screenshot 2019-04-25 at 12 57 13
Later versions of Safari by comparison:
Screenshot 2019-04-25 at 12 58 31

One of the tests periodically returns a deadline > 50ms in Safari 9 so I've added some defensive code to prevent that. It's frustrating I haven't been able to get to the bottom of this completely.

notlee:
The tests are failing in Firefox as `requestAnimationFrame` is not always fired consistently during the test. I'm unsure why this is. More unusually, I can't replicate when viewing in browser stack live and if I just run the tests through firefox they appear to work:

Screenshot 2019-04-26 at 11 37 35

They're more likely to fail when more browsers are tested, but this feels like a false lead:
Screenshot 2019-04-25 at 21 38 49

polyfills/requestIdleCallback/polyfill.js
```diff
@@ -0,0 +1,306 @@
...
+        return Math.min(frameDeadline - performance.now(), 50);
+    };
+
+    function getDeadline(callbackObject) {
+        var timeout = callbackObject.options.timeout;
+        var added = callbackObject.added;
+        // Create deadline from global.IdleDeadline constructor throws a type
+        // error. So create a new object which inherits its prototype.
+        var deadlinePrototype = Object.create(global.IdleDeadline.prototype);
+        function IdleDeadline() {};
```
JakeChampion: I think this is meant to `throw new TypeError('Illegal constructor')`

polyfills/requestIdleCallback/polyfill.js
```diff
@@ -0,0 +1,306 @@
...
+        isCallbackRunning = true;
+
+        // Move timed-out callbacks from the scheduled array.
+        timedOutCallbacks = scheduledCallbacks.filter(function (callbackObject) {
+            if (callbackObject.options.timeout !== undefined) {
+                var deadline = getDeadline(callbackObject);
+                return deadline.didTimeout;
+            }
+            return false;
+        }, []);
```
JakeChampion: I don't think we need to set the `thisArg` of Array.prototype.filter to an array.
```suggestion
        });
```

polyfills/requestIdleCallback/polyfill.js
JakeChampion: This is done further down in the polyfill, which I think is incorrect because it would mean that `requestIdleCallback((deadline) => console.log(deadline instanceof IdleDeadline))` would log out `false` when it should log out `true`

notlee: The [global IdleDeadline polyfill](https://github.com/Financial-Times/polyfill-library/pull/192/files/6e13d7ca54429c15ca9953c822d5c393528d17d2#diff-99660d69ae7687cb56494acf81bb7c34R284) does throw a type error in its constructor. So I can create an instance to return this function inherits its prototype. Is that an odd approach?

JakeChampion: Something like this should work:
```js
var throwError = false;
function IdleDeadline() { if(throwError) throw new TypeError('Illegal constructor')}
var deadlinePrototype = Object.create(global.IdleDeadline.prototype);
IdleDeadline.prototype = deadlinePrototype;
var deadline = new IdleDeadline();
throwError = true;
```

polyfills/requestIdleCallback/tests.js
```diff
@@ -0,0 +1,360 @@
...
+
+    it('schedules a callback with one IdleDeadline argument passed to the callback', function (done) {
+        requestIdleCallback(function (deadline) {
+            proclaim.equal(
+                arguments.length,
+                1,
+                'Expected the callback to receive one argument.'
+            );
+
+            proclaim.isTrue(deadline instanceof IdleDeadline);
```
JakeChampion: 👍

notlee:
Jake and I discussed the [simplifying IdleDeadline](https://github.com/Financial-Times/polyfill-library/pull/192/commits/5501c08cf91d51927bd3afd54ba8301608249ebe). :shipit:

PR #193, opened 2019-04-26
({}).toString.call(null) should not be [object Window]

### LINKED ISSUE #164: es6 and bootstrap 4.3.1 have issues on IE11
I've recently updated one of my website using polyfill.io to bootstrap 4.3.1 and now tooltips cause js error on IE11. The error goes away when removing polyfill.io or downgrading bootstrap to 4.3.0.

The issue seems linked to the "es6" bundle (if I only add es5, it works fine). Not sure which feature is causing the error, I'm still looking into it.

Bootstrap 4 issue for reference: https://github.com/twbs/bootstrap/issues/28520

EamonNerbonne:
Fixes #164

### DIFF
--- a/polyfills/Symbol/polyfill.js
+++ b/polyfills/Symbol/polyfill.js
@@ -213,6 +213,7 @@
 	defineProperty(Object, 'create', descriptor);
 
 	descriptor.value = function () {
+		'use strict'; //so that ({}).toString.call(null) returns the correct [object Null] rather than [object Window]
 		var str = toString.call(this);
 		return (str === '[object String]' && onlySymbols(this)) ? '[object Symbol]' : str;
 	};
--- a/polyfills/Symbol/polyfill.js
+++ b/polyfills/Symbol/polyfill.js
@@ -20,7 +20,7 @@
 	var pIE = ObjectProto[PIE];
 	var toString = ObjectProto.toString;
 	var concat = Array.prototype.concat;
-	var cachedWindowNames = typeof window === 'object' ? Object.getOwnPropertyNames(window) : [];
+	var cachedWindowNames = Object.getOwnPropertyNames ? Object.getOwnPropertyNames(window) : [];
 	var nGOPN = Object[GOPN];
 	var gOPN = function getOwnPropertyNames (obj) {
 		if (toString.call(obj) === '[object Window]') {
--- a/polyfills/Symbol/polyfill.js
+++ b/polyfills/Symbol/polyfill.js
@@ -220,7 +220,7 @@
 			return (str === '[object String]' && onlySymbols(this)) ? '[object Symbol]' : str;
 		};
 	} else {
-		descriptor.value = toString;
+		descriptor.value = toString.bind(window);
 		defineProperty(window, 'toString', descriptor);
 		descriptor.value = function () {
 			// https://github.com/Financial-Times/polyfill-library/issues/164#issuecomment-486965300
--- a/polyfills/Symbol/polyfill.js
+++ b/polyfills/Symbol/polyfill.js
@@ -2,6 +2,7 @@
 // (C) Andrea Giammarchi - MIT Licensed
 
 (function (Object, GOPS, global) {
+	'use strict'; //so that ({}).toString.call(null) returns the correct [object Null] rather than [object Window]
 
 	var	setDescriptor;
 	var id = 0;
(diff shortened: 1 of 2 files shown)


EamonNerbonne:
Works on my machine (whatever that's worth!):

image

EamonNerbonne:
@JakeChampion So, IE8 and 9 don't support use strict, so there's a choice here: either don't support this corner case for them (and update the test to e.g. not assert if docMode <=9); or try and support a hacky hacky workaround when we can detect that 'use strict' didn't work.

I did the latter here, but I can image it's not worth the complexity.

polyfills/Symbol/polyfill.js
```diff
@@ -212,13 +213,25 @@
...
-		return (str === '[object String]' && onlySymbols(this)) ? '[object Symbol]' : str;
-	};
-	defineProperty(ObjectProto, 'toString', descriptor);
+	var strictModeUnsupported = (function(){ return this; }).call(null) !== null;
+	if(strictModeUnsupported) {
+		descriptor.value = toString;
+		defineProperty(window, 'toString', descriptor);
+		descriptor.value = function () {
+			if(this === window) {
+				return '[object Null]';
```
JakeChampion: I don't think we can allow this because `Object.prototype.toString.call(window)` should return `[object Window]`

polyfills/Symbol/tests.js
```diff
@@ -58,8 +58,9 @@ it('should return "[object Symbol]" when called with Object.prototype.toString()
 	proclaim.equal(Object.prototype.toString.call(Symbol()), '[object Symbol]');
 });
 
-it('should retain browser toString behavior for nulls', function() {
+it('should retain browser toString behavior for nulls and window', function() {
 	proclaim.equal(Object.prototype.toString.call(null), '[object Null]');
+	proclaim.notEqual(window.toString(), '[object Null]');
```
JakeChampion: We need a test for `proclaim.equal(Object.prototype.toString.call(window), '[object Window]')`

EamonNerbonne:
> We need to allow `Object.prototype.toString.call(window)` to return `[object Window]`

On most browsers, this will.  But on IE9+8 there's no way to distinguish between null and window when toString is implemented in javascript, i.e. `Object.prototype.toString.call(window) === Object.prototype.toString.call(null)`.

So you *have* to choose to make a mistake.  This PR chooses to keep `Object.prototype.toString.call(null) === '[object Null]'`.  I'll make an alternative picking the other option, but both are necessarily incorrect.

In terms of compat impact, I think keeping `null` correct is more important, firstly because it's more like a "special" value, and secondly, because IE8 and IE9 actually differ in what they return: IE8 will return `[object Object]` regardless, so the details probably don't matter, and thirdly, code that calls toString directly on the object, (i.e. not via `Object.prototype`) *can* be fixed.

It appears IE8 returns `[object Object]` anyhow, for both window and null, so for IE8 I doubt it matters 
whether `Object.prototype.toString.call(window)=== '[object Window]'

Anyhow, that's why this PR is as it is.  I'll make a second one for the alternative, but there is (as far as I can tell) no correct solution.  Existing behavior is to have correct behavior for window, but not for null, even on IE10 and newer.

EamonNerbonne:
As far as I can tell, IE8+9 market share is pretty much nil, and they're unsupported by MS, which makes either PR equivalent, and this one is simpler.

polyfills/Symbol/tests.js
EamonNerbonne: Test added; but for ancient IE this isn't true.

EamonNerbonne:
@JakeChampion just a quick ping - have you had time to consider which solution is least bad?  The current situation on master is in any case worse, since it breaks on all modern browsers; #198 is a clean fix in modern browsers and IE10 or newer (but simply doesn't affect IE8+9 at all), and this PR is a fix for modern browsers and IE10 or newer, and tries to pick a not-too-bad compromise for ancient IE (but will break some corner cases on those browsers).

polyfills/Symbol/tests.js
```diff
@@ -30,6 +30,7 @@ var arePropertyDescriptorsSupported = function () {
 	}
 };
 var supportsDescriptors = Object.defineProperty && arePropertyDescriptorsSupported();
+var strictModeSupported = (function(){ return this; }).call(null) === null;
```
EamonNerbonne: I don't know enough about the polyfill pipeline, but is this file implicitly executed in strict mode?  (After all,  polyfills/Symbol/polyfill.js clearly is not.)

This would be less ambigous as a test reader using something like this:
```suggestion
var strictModeSupported = (function(){ 'use strict'; return this; }).call(null) === null;
```

JakeChampion: Good idea! I've added this in https://github.com/Financial-Times/polyfill-library/pull/193/commits/7fdae22972c88346bdd75b9c76b03389d68efa2f

polyfills/Symbol/polyfill.js
```diff
@@ -20,7 +20,7 @@
 	var pIE = ObjectProto[PIE];
 	var toString = ObjectProto.toString;
 	var concat = Array.prototype.concat;
-	var cachedWindowNames = typeof window === 'object' ? Object.getOwnPropertyNames(window) : [];
```
EamonNerbonne: so I can reproduce the circleci reported test failure in IE11 compat IE8 mode on win10.

EamonNerbonne:
The failing test is because in IE8 window doesn't quite work as it does in other browsers.  I don't believe it's actually related to the symbol polyfill at all; the test was merely added to ensure that for modern browsers the polyfill doesn't break existing behavior.

(i.e. the aim of the test was to ensure the polyfill doesn't break unrelated stuff, not to actually polyfill window.toString()).

The test should simply be disabled for IE8.

(Aside: is the added complexity and bug-proneness and size of code that needs to be IE8 compatible really worth it given that browsers market share?)

EamonNerbonne:
An alternative incidentally (which I really think is kind of pointless given how niche it is) would be to assign something like this `window.toString = function() { return "[object Window]"; };` which won't help the general case that's abusing Object.prototype.toString to do type-detection as bootstrap is, but would help direct callers.

EamonNerbonne:
(I'd definitely prefer to keep that tweak out of this PR, given the surprising complexity this PR already turned out to have)

PR #198, opened 2019-05-02
({}).toString.call(null) should not be [object Window] (alternative)

### LINKED ISSUE #164: es6 and bootstrap 4.3.1 have issues on IE11

EamonNerbonne:
Alternative to #193; fixes #164.

For old-IE, there's no use strict, and this PR accepts the current behavior of that `Object.prototype.toString.call(null)` (and `Object.prototype.toString.call(window)`) both return `[object Window]` on IE9, and `[object Object]` on IE8.  For IE10, IE11, and modern browsers this PR and #193 behave identically: `Object.prototype.toString.call(window) === '[object Window]'` and `Object.prototype.toString.call(null) === '[object Null]'`.

In essence: for IE8+9 the Symbol polyfill necessarily introduces a bug into toString.  Current behavior makes `null` wrong.  This PR keeps that bug; the alternative #193 fixed the `null` argument, but breaks `window`.  But either PR fixes the bug for IE10 and newer.

### DIFF
--- a/polyfills/Symbol/polyfill.js
+++ b/polyfills/Symbol/polyfill.js
@@ -213,6 +213,7 @@
 	defineProperty(Object, 'create', descriptor);
 
 	descriptor.value = function () {
+		'use strict'; //so that ({}).toString.call(null) returns the correct [object Null] rather than [object Window]
 		var str = toString.call(this);
 		return (str === '[object String]' && onlySymbols(this)) ? '[object Symbol]' : str;
 	};
--- a/polyfills/Symbol/polyfill.js
+++ b/polyfills/Symbol/polyfill.js
@@ -2,6 +2,7 @@
 // (C) Andrea Giammarchi - MIT Licensed
 
 (function (Object, GOPS, global) {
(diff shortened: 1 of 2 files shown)


EamonNerbonne:
@JakeChampion this better?

EamonNerbonne:
@JakeChampion you're probably busy, but it'd be great if there's at least some solution for #164 - I'm not sure if you've just not been able to get around to this or whether there's an objection?  Some clarity would be great...

EamonNerbonne:
*ping*?

JakeChampion:
We will go with your other solution 👍 https://github.com/Financial-Times/polyfill-library/pull/193

PR #204, opened 2019-05-14
Fix 'Promise.prototype.finally' IE version

ku00:
I found a problem. 
It should be `<12` instead of `>12`, shouldn't it?


JakeChampion:
Thanks for the contribution @ku00. It should be `>12`, this is because Internet Explorer does not have `Promise` at all, so it would get served out `Promise` polyfill which already has a `Promise.prototype.finally` method. This separate polyfill for `Promise.prototype.finally` is for browsers which have a native `Promise` that does not have the `Promise.prototype.finally` method.

Issue #225, opened 2019-06-21
Bug: Map detection fails in strict mode


Ruud-Zuiderlicht:
## What
I noticed Map is being polyfilled even though I have it gated. The browser is Chrome 75 (latest release at the moment). I did bypass the browser detection by downloading the actual code using IE. This was needed to make sure 3rd parties have the polyfills.

It turns out the browser runs into a feature-detect check that errors out. The check in question is `function () { ... 'Symbol' in this ... }`. Where this is `undefined`. This makes the code execute the catch clause. (See https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Map/detect.js)

## Details
The expected behavior is that Map isn't polyfilled, because the detection script skips this part of the code.

The actual behavior is that Map is being polyfilled.

To reproduce,
1. Include the generated code from polyfill.io using these settings: https://polyfill.io/v3/polyfill.js?flags=gated&features=Map (Chrome will generate no code, I had to download using IE to include the code)
2. Open the page containing step 1. there will be no error because of a try .. catch
3. Place a break point at this line https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Map/detect.js#L6
4. Inspect `this`; it wil be undefined

Browser: Google Chrome is up to date, Version 75.0.3770.100
Device: Desktop
![image](https://user-images.githubusercontent.com/18439859/59908480-d0cb9780-940d-11e9-9cf3-f7a26950098d.png)

_edit: I've updated the title as this issue isn't related to a bug in Chrome, but the usage of strict mode outside of the polyfill code._

JakeChampion:
Hi there @Ruud-Zuiderlicht.

Which version of IE did you use to generate the bundle? When I use IE/11 and run the polyfill bundle in Chrome, it works correctly.

URL I used: https://polyfill.io/v3/polyfill.js?flags=gated&features=Map&ua=ie/11
Here is a JSBin which shows it working as expected: https://jsbin.com/codetezuye/edit?html,console

Ruud-Zuiderlicht:
I tried the JSBin and didn't get the error either, but I forgot to mention that I placed the code inside a file that is compiled by Webpack. I tried moving the polyfill outside of that file and then everything works as expected. So I expect Webpack to declare strict mode, which explains the behavior. I've moved the code outside of Webpack to fix for now.

Is there a way to improve on this on your side as well?

JakeChampion:
[...]

PR #226, opened 2019-06-24
Make entire bundle in strict mode

JakeChampion:
Attempting to see if this can resolve https://github.com/Financial-Times/polyfill-library/issues/225

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -300,7 +300,7 @@ function getPolyfillString(opts) {
 				// Outer closure hides private features from global scope
 				output.add(streamFromString("(function(undefined) {" + lf));
 				// Strict mode for the entire bundle
-				output.add(streamFromString("use strict"));
+				output.add(streamFromString("'use strict';" + lf));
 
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -299,6 +299,8 @@ function getPolyfillString(opts) {
 			if (sortedFeatures.length) {
 				// Outer closure hides private features from global scope
 				output.add(streamFromString("(function(undefined) {" + lf));
+				// Strict mode for the entire bundle
+				output.add(streamFromString("use strict"));
```
Ruud-Zuiderlicht: This breaks the bundle, there should be quotes around "use strict". Is this the correct fix to add those?

`output.add(streamFromString("'use strict'" + lf));`

Ruud-Zuiderlicht:
This fix invokes the problem itself. I had thought to change the detection script (see [498ab55](https://github.com/Financial-Times/polyfill-library/commit/498ab55dea090b207f85d812823b562647073c09)). But I'm not sure this will work as expected on all of the target browsers.

PR #233, opened 2019-06-24
Fix broken this reference anonymous function when in strict mode

### LINKED ISSUE #225: Bug: Map detection fails in strict mode

JakeChampion:
Resolves #225

### DIFF
--- a/polyfills/Map/detect.js
+++ b/polyfills/Map/detect.js
@@ -1,9 +1,9 @@
-'Map' in this && (function() {
+'Map' in this && (function(self) {
 	try {
 		var m = new Map([[1, 1], [2, 2]]);
 		if (Map.length === 0) {
 			if (m.size === 2) {
-				if ('Symbol' in this && 'iterator' in Symbol && typeof m[Symbol.iterator] === 'function') {
+				if ('Symbol' in self && 'iterator' in Symbol && typeof m[Symbol.iterator] === 'function') {
 					return true;
 				}
 			}
@@ -12,4 +12,4 @@
 	} catch (e) {
 		return false;
 	}
-}())
+}(this))


JakeChampion:
@Ruud-Zuiderlicht, I've added your commit to improve the feature detect for Map. If you are happy for this commit to be added to the project I would need you to accept our [Contributor License Agreement](https://cla-assistant.io/Financial-Times/polyfill-library?pullRequest=233). If you are not happy to accept the agreement, let me know and I can close this pull-request :-)

PR #251, opened 2019-06-25
add HTMLTemplateElement as an available polyfill

nicksrandall:
With the growing popularity of UI libraries like `lit-html` and `hyperHTML` which use HTMLTemplateElement under-the-hood, I think it makes sense to include a minimal polyfill for that here.

One potential concern I have is that the polyfill author does not ship a minified version of the library in their npm module. I assume that means that we will have to either minify it ourselves or serve the polyfill un-minified.

### DIFF
--- a/polyfills/HTMLTemplateElement/config.json
+++ b/polyfills/HTMLTemplateElement/config.json
@@ -1,14 +1,14 @@
 {
   "browsers": {
-    "edge": "<12",
-    "edge_mob": "<12",
+    "edge": "<13",
+    "edge_mob": "<13",
     "ie": "11",
     "ie_mob": "11",
     "firefox": "4 - 22",
     "opera": "10 - 22",
-    "safari": "6 - 9",
-    "chrome": "* - 35",
-    "ios_saf": "* - 9",
+    "safari": "6 - 8",
+    "chrome": "* - 26",
+    "ios_saf": "* - 8",
     "firefox_mob": "<67",
     "bb": "7 - 10",
     "android": "<67"
--- a/polyfills/HTMLTemplateElement/config.json
+++ b/polyfills/HTMLTemplateElement/config.json
@@ -1,0 +1,32 @@
+{
+  "browsers": {
+    "edge": "<12",
+    "edge_mob": "<12",
+    "ie": "9 - *",
+    "ie_mob": "10 - *",
+    "firefox": "4 - 22",
+    "opera": "10 - 22",
+    "safari": "6 - 9",
+    "chrome": "* - 35",
+    "ios_saf": "* - 9",
+    "firefox_mob": "<67",
+    "bb": "7 - 10",
+    "android": "<67"
+  },
+  "dependencies": [
+    "DocumentFragment",
+    "document.querySelector",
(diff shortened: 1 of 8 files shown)


polyfills/HTMLTemplateElement/config.json
```diff
@@ -0,0 +1,25 @@
...
+		"firefox": "4 - 22",
+		"opera": "10 - 22",
+		"safari": "6 - 9",
+		"chrome": "* - 35",
+		"ios_saf": "* - 9",
+		"firefox_mob": "<67",
+		"bb": "7 - 10",
+		"android": "<67"
+	},
+	"dependencies": ["~html5-elements", "DocumentFragment"],
```
JakeChampion: This polyfill does not depend on ~html5-elements as far as I can see.
This polyfill does depend on:
- DocumentFragment
- querySelectorAll
- querySelector
- Object.defineProperties
- Object.getOwnPropertyDescriptor
- Object.defineProperty
- document.addEventListener

polyfills/HTMLTemplateElement/polyfill.js
```diff
@@ -0,0 +1,600 @@
+/**
```
JakeChampion: We would need to add this filepath to .gitignore and .eslintignore

polyfills/HTMLTemplateElement/config.json
```diff
@@ -0,0 +1,25 @@
+{
+	"browsers": {
+		"edge": "<12",
+		"edge_mob": "<12",
+		"ie": "*",
```
JakeChampion: Looking at the polyfill, I don't think it supports Internet Explorer 8 due to using getters, I may be wrong though.

JakeChampion:
@nicksrandall Thank you for the pull-request, it is looking good so far.

Would it be possible for you to add some tests for the polyfill to ensure it works in all the browsers we have configured it to be served to? I'm really interested to see if this works in Internet Explorer 8, 9, and 10.

nicksrandall:
@JakeChampion I added some tests but I only have the ability to test on IE 11 (everything worked). Are you able to run the tests on any older browsers? 

I also added the dependencies that you listed.

nicksrandall:
I'm not sure what to do here... It looks like the polyfill only works for IE 11. Should I update the config.json to only include the polyfill for IE 11 since that is the only place that it is helpful?

JakeChampion:
@nicksrandall Yep, I agree, I've updated the config to only serve IE 11 but can see that the tests fail in the Chrome versions that the polyfill is configured to serve, is that something you could look into at all?

nicksrandall:
@JakeChampion I updated the config to be more specific on the versions where the polyfill works and is needed. We'll see how the next CI round goes.

nicksrandall:
@JakeChampion I'm having a hard time reading these CI test results. I can't tell why it failed?

JakeChampion:
Thanks for this contribution @nicksrandall!

nicksrandall:
Would it be helpful if I created a PR to update the version of this library in polyfill-service?

Issue #219, opened 2019-06-18
`Map` polyfill (`get(key)`) is very slow 


craigkovatch:
We recently starting using `polyfill-library@3.34.0` to polyfill ES6 for (only) IE11 in Tableau. This caused a severe performance regression in our rendering code, which was traced to the `Map` polyfill. Removing `Map` and `Set` from our config resolved the issue -- luckily our code only currently needs the native partial support for `Map` in IE11. (I don't know if removing `Set` was also necessary, but since IE11 has partial support for both, I decided to err on the side of removing both.)

Unfortunately I have not investigated deeply enough to suggest a fix, but our profiling session indicated `CreateMethodProperty(Map.prototype...) (23.1.3.6 Map.prototype.get (key))` was the culprit.

Here is our config, for reference:
```
const ie11_es6_opts = {
  uaString: 'Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; rv:11.0) like Gecko',
  minify: false,
  excludes: [
    'Map',
    'Set',
  ],
  features: {
    'es6': { flags: ['gated'] }
  }
};
```

neenhouse:
Observed same perf issue with Map.  `SameValueZero` is executed millions of times in comparison to `get` and `set` being executed 10s of thousands of times.

neenhouse:
I suspect the root of the perf issue is mainly the internal algo's `Map` uses are O(n) computations, and when you are performing another O(n) operation on a Map structure the perf cost hit becomes O(n^2) and you get exponential slow down.  See https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Map/polyfill.js#L272 for examples.

Solution could be to make the iteration computation O(1) for Map's internals using some kind of hash table or maybe even converting the internal `_keys` property to an object (instead of an array).   I could get the ball rolling on this idea.  This issue is definitely hindering us from fully leveraging these polyfills.

JakeChampion:
@neenhouse If that optimisation can be made whilst adhering to the specification steps, we will gladly accept the contribution. Thank you for offering to work on this :-)

zloirock:
> If that optimisation can be made whilst adhering to the specification steps

Specification steps are not something that should be interpreted as an algorithm for implementation. [See the spec](https://tc39.es/ecma262/#sec-map-objects):

[...]

PR #293, opened 2019-08-14
#219 - Improve Map performance

neenhouse:
Addresses #219 

Only has test plan.  Implementation still pending.


neenhouse:
First commit only adds a performance unit test that consistently fails IE11 locally (per contributing.md)

JakeChampion:
@neenhouse Thanks for working on this, I've triggered a run on CircleCI for this. 

In order for us to be able to accept your contribution though, we require you to have read and accepted our [contributor license agreement](https://cla-assistant.io/Financial-Times/polyfill-library?pullRequest=293).

neenhouse:
Test fails in CI in expected way :+1:

neenhouse:
Took a first pass at the algorithm.  Two notes:

- I avoided climbing the ladder of abstraction for first pass, so some code (normalize key + retrieve from hash table) is not DRY yet.
- When looking up functions and objects by key, algo currently assigns an internal id property to the object `__meta__.id` so it can identify the object in the internal hash table.  There are alternatives to this approach that don't involve modifying the key function or object, but they won't be as performant. 

I'd like to run a few more tests to compare performance between master + branch.  Also want to try at least one algo that does not do the key modification approach and instead uses smaller hash tables that are iterated on to handle those lookups as edge cases.

neenhouse:
Running local Karma unit test in Chrome `76.0.3809.100 (Official Build) (64-bit)` comparing native implementation of `Map` vs polyfill-library:

- Native Chrome: 86ms, 107ms, 110ms, 94ms, 96ms, 81ms, 107ms, 100ms
- Polyfill Library (branch): 127ms, 116ms, 124ms, 116ms, 128ms, 116ms, 99ms, 121ms
- Polyfill Library (master): fails consistently (after 1000ms when Map size reaches around 6000 / 10000 records)

Screen Shot 2019-08-15 at 8 37 47 AM

JakeChampion:
Looking great @neenhouse!

neenhouse:
@JakeChampion Think this is ready for 1st review.  I improved the marking of objects/functions by introducing randomness to avoid collisions if lib is loaded multiple times and abuse.  Using reference comparison based algos even in smaller hash tables eventually leads to the same O(n) perf problem if a particular key type is over-used, so I thought this would be an acceptable trade off (thought also since primitives are usually used as keys in most UIs I've worked on, maybe a moot issue?).

zloirock:
- Metadata should be unobservable (`Object.assign`, `Object.getOwnPropertySymbol`, `Reflect.ownKeys`, etc. on keys).
- It should work with frozen objects.
- `.size` and `.delete` also could be `O(1)`.

neenhouse:
@zloirock Thanks for the feedback.  Frozen object keys forced me to climb the ladder of abstraction by 1 rung and were a bit tricky.  I don't see a way to avoid O(n) computation time when dealing with those.  Do you have any suggestions for dealing with perf in those cases?  Appreciate any additional feedback.

@JakeChampion - I saw for `size` you already were precomputing a `_size` property and thought it would be a good idea (as @zloirock suggested) to convert the `size` getter to implement that instead of performing a O(n) iteration.  I couldn't spot any logical fallacies for doing that.  Is there anything I'm missing there?

zloirock:
@neenhouse see `core-js`.

neenhouse:
Thanks for the feedback @zloirock

neenhouse:
☝️ Last CI failure looks like browser stack flake:

Screen Shot 2019-08-16 at 2 24 08 PM

craigkovatch:
Will this also fix the perf issues in Set? I'm guessing they have the same root cause.

neenhouse:
Hi @craigkovatch, this only improves `Map`, but `Set` has the same underlying issue with O(n) perf.  I would imagine the right path would to be either to:

- Abstract a private class to use between Map and Set that implement similar, optimized operations.
- Re-implement the same optimizations used in Map in Set.

craigkovatch:
Any update on this? We would really like to be able to use Set/Map

PR #295, opened 2019-08-17
#294 - Opt IE11 into `Object.keys` polyfill

### LINKED ISSUE #294: [feature-request] `Object.keys` should coerce primitive values in IE11
## Overview 

In IE11, running the following causes a uncontrolled exception to occur:

```
Object.keys(50);
```

This is problematic for applications written with modern JS, babel is typically used to allow statements like:

```
const myData = {...dataA, ...dataB}
```

Babel will transpile that line and implement a function called `_objectSpread`: https://github.com/babel/babel/blob/a112be2fa0b2abab7fd9a845b3a76b273c506449/packages/babel-helpers/src/helpers.js#L393, which implements `Object.keys`:

```
...
var ownKeys = Object.keys(source);
...
```

This is a highly traveled path in UI apps that implement object spread syntax (which is pretty common these days).  Spreading non object type data is a pretty easy mistake to make, and modern browsers forgive.  

To prevent exception in IE11 with string and number, we should use the polyfill provided from this lib which already follows the ECMA-262 ed 6 spec (http://ecma-international.org/ecma-262/6.0/index.html#sec-object.keys).

## Why does this matter

Companies who are trying to optimize their polyfill payload can very likely run into this issue when they try to swap out core-js `Object` APIs for polyfill-library with apps that implement object spread.  This issue can cause complete failure in apps and prevent adoption of polyfill-library.  

## Browsers impacted

- IE11
- Others?

I should be able to prepare a PR pretty soon to address this issue.

neenhouse:
Closes #294


neenhouse:
Created first commit to modify unit tests to change:

- undefined, null, and empty arguments should return empty Array.

neenhouse:
Digging into the issue a little further (and the spec), I saw this: http://ecma-international.org/ecma-262/6.0/index.html#sec-toobject

So I think there really isn't any need to change the test plan since the assertions match what the spec needs.  I think this issue could be solved by a small tweak to the polyfill detection.

Will force push out a new commit to rollback test commit.

JamieMason:
FYI the build has failed on test_11 but it looks to be just a timeout in Browserstack.

<details>

```
#!/bin/bash -eo pipefail
node ./test/polyfills/test-individual-feature.js Element.prototype.inert

Running tests for Element.prototype.inert because one or more of the files it depends on has changed.
The files which changed were polyfills/Object/keys/config.toml,polyfills/Object/keys/detect.js
Testing Element.prototype.inert

START:
20 08 2019 15:41:27.602:WARN [launcher.browserstack]: chrome 44.0 (Windows XP) has not captured in 60000 ms, killing.
20 08 2019 15:41:27.604:WARN [launcher.browserstack]: chrome 45.0 (Windows XP) has not captured in 60000 ms, killing.
20 08 2019 15:46:49.787:WARN [launcher.browserstack]: edge 17.0 (Windows 10) has not captured in 60000 ms, killing.
20 08 2019 15:46:55.517:WARN [launcher.browserstack]: edge 18.0 (Windows 10) has not captured in 60000 ms, killing.
20 08 2019 15:49:19.962:WARN [launcher.browserstack]: edge 17.0 (Windows 10) has not captured in 60000 ms, killing.
20 08 2019 15:50:53.412:ERROR [reporter.browserlabs]: ✖ Could not update BrowserStack status
Too long with no output (exceeded 3h0m0s)
```

</details>

oberhamberger:
> FYI the build has failed on test_11 but it looks to be just a timeout in Browserstack.

Is there a way to restart the build somehow?

oberhamberger:
@JakeChampion tests are still failing :(
this time on test_4

IsabellaZaby:
Could anyone retrigger the build? test_11 failed this time...

JakeChampion:
@IsabellaZaby There's no need to trigger a rebuild on CI because the tests failed due to getting results that are incorrect, we only need to trigger a rebuild when BrowserStack has timed out.

DullReferenceException:
Apparently `Object.keys('')` also fails in IE11 as well, so the additional check may not be necessary. Created this PR before noticing this one: https://github.com/Financial-Times/polyfill-library/pull/317

PR #301, opened 2019-09-03
Adds support for Intl.RelativeTimeFormat and Intl.PluralRules

TheDadi:
Adds support for:
- Intl.RelativeTimeFormat
- Intl.PluralRules
and their locale files


JakeChampion:
@TheDadi Thank you for taking the time to contribute :-) Are you able to add some tests for these changes, it would be great to know that those polyfills work in all the browsers we have configured them to be served to.

TheDadi:
will do so

TheDadi:
@JakeChampion They are not as easy to test. Because i need to be able to import a locale file in to the test and also be able to make some deps on other polyfills they need to work. Because the Intl.RelativeTimeFormat needs Intl, Intl.NumberFormat and Intl.PluralRules to be available to work. How do you suggest to implement a meaningful test without changing the current test setup?

TheDadi:
@JakeChampion added some test's let me know if your fine with them.

TheDadi:
bump

PR #309, opened 2019-10-23
Updating Symbol.iterator

hewholived:
Updating Symbol.iterator to capture Symbol from global rather than capturing it ambiently


JakeChampion:
@hewholived Does the current version of the polyfill not work in a scenario you have come across?

hewholived:
Sometimes it is useful to execute polyfills in a non side-effecting manner. For example, if the polyfill is used in a 3rd party library of some sort, you would not want to mutate the global object and attach polyfills to it. Using self.Symbol instead of Symbol will ensure that Symbol is accessed from a self object (which could be an object which is not a real window/global). @JakeChampion

PR #314, opened 2019-10-30
Bind `this` in methods that use it

salemhilal:
## what's going on
Polyfill helpers use `this` to reference the global scope, which is true at declaration time since the generated polyfill code explicitly wraps everything in an iife that binds the global scope to `this`.

However, if this code is later executed in a strict-mode context, it would fail, since `this` is whatever the parent's context is. We experienced this when testing a static version of generated polyfill code, which inserted the code into a strict-mode-enabled wrapper.

## what's the fix
This commit makes four methods strict-mode compliant by binding `this` to its value at declaration time, which is set to the global scope.

## how did you test it
I tested this by ensuring `npm run test` passed (the klarna part fails, but I manually ran it against the features we ship using `npm run test-polyfills -- --features=Array.from,Array.of,...`). I also validated that shipping this patch on a static version of a polyfill file fixed strict-mode errors we saw in production at Etsy.


polyfills/_ESAbstract/ArraySpeciesCreate/polyfill.js
```diff
@@ -45,4 +45,4 @@ function ArraySpeciesCreate(originalArray, length) { // eslint-disable-line no-u
 	}
 	// 10. Return ? Construct(C, « length »).
 	return Construct(C, [length]);
-}
+}).bind(this);
```
JakeChampion: Internet Explorer 8 does not support the `Function.prototype.bind` method.

I think instead of using the context (`this`) as the reference to the global object, we should instead change the polyfill-library wrapper code to have the global object be an argument.

salemhilal: oh wow, good catch, and I like that idea. I'll do some more digging and commit my updates here.

JakeChampion: I had an attempt at updating the wrapper code and polyfills on this branch --> https://github.com/Financial-Times/polyfill-library/compare/self?expand=1

salemhilal: This is perfect. It's a lot to look over but the detection.js changes look good, and the _ESAbstract changes look good as well. For what it's worth, I only had to change those `function Name()` calls to `var Name = function()` in order to allow those `.bind()` calls to work — leaving those as named functions or not is up to you

JakeChampion:
This was superseded by https://github.com/Financial-Times/polyfill-library/pull/318

salemhilal:
Thank you!

Issue #297, opened 2019-08-21
URL polyfill fails in IE if used with strict mode bundling process


niksy:
## What

When trying to use `URL` polyfill with bundling JS with bundler like Webpack or Rollup, polyfilling fails in IE11 if module is in strict mode (using it with ES `import` makes code behave in strict mode). This is related to [already closed issue](https://github.com/Financial-Times/polyfill-library/issues/225), but now error happens in different part of codebase, pinpointed [here](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/_ESAbstract/Type/polyfill.js#L18).

It seems like referencing `this` makes it `undefined` in strict mode. Should this be changed to something like `self`? Also, if this is welcome change, maybe this should be reflected to every part of codebase where `this.Symbol` is defined.

## Details

Polyfills are expected to work in strict mode when `this` is used to reference globally defined properties such as `Symbol`.

JakeChampion:
Thanks for the issue, I agree with changing `this` to `self`, will you be making a pull-request with the change? If not, I will happily make the change myself :-)

niksy:
I will make PR!

niksy:
@JakeChampion done!

PR #318, opened 2019-11-13
switch the global object from being `this` to being an argument named `self`

JakeChampion:
This enables the bundle to work in strict mode, which some of our users are trying to execute the bundle in.

Solves #297 
Solves #298
Solves #314

This work should hopefully enable us to merge in #226


notlee:
🤞 you can do it browserstack!

niksy:
When can we expect this to be merged? Anything we can help with?

notlee:
👍 This time!? I still approve; I approve more than ever

jimmywarting:
hmm, maybe a bit late to the party but shouldn't it have been [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) instead?

JakeChampion:
@jimmywarting It will be eventually, yes :-)

PR #347, opened 2020-01-10
Ensure all code which runs in a browser environment is written in ES3 syntax

JakeChampion:
This avoids potential bugs slipping in such as usinng keywords in dot-notation, which causes Internet Explorer 8 to throw an error during the parsing of the JavaScript.

### DIFF
--- a/polyfills/.eslintrc
+++ b/polyfills/.eslintrc
@@ -1,43 +1,19 @@
 {
-	"env": {
-		"es6": false,
-		"browser": true,
-		"node": false
-	},
-	"parserOptions": {
-		"ecmaVersion": 3
-	},
-	"rules": {
-		"no-unused-vars": 2,
-		"no-undef": 1,
-		"eqeqeq": 0,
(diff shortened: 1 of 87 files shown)


polyfills/.eslintrc
```diff
@@ -1,43 +1,19 @@
...
+  "parserOptions": {
+    "ecmaVersion": 3,
+    "sourceType": "script"
+  },
+  "rules": {
+	"comma-dangle"          : [ 2, "never" ],
+	"quote-props"           : [ 2, "as-needed", { "keywords" : true } ],
+	"dot-notation"          : [ 2, { "allowKeywords" : false } ],
+	"radix"                 : 2,
+	"no-catch-shadow"       : 2
```
notlee: >In IE 8 and earlier, the catch clause parameter can overwrite the value of a variable in the outer scope, if that variable has the same name as the catch clause parameter.

:mind-blown:

PR #392, opened 2020-01-22
Author polyfill configurations in TOML but publish in JSON to avoid consumers requiring a TOML parser

JakeChampion:
This allows the us to drop TOML from dependencies section and gets us closer to getting this package working in browser and cloudflare worker environments with a smaller bundle size


JakeChampion:
If the first polyfill's test suite passes, we can merge this without waiting for the other tests to run because if this was to break it would break all the polyfills not just some of them.

cc @chee

Issue #177, opened 2019-03-30
Use Async/Await to improve performance


underbyte:
## Details
Any investigation done in switching over all the promises over to async/await implementation? I'm interested if switching would improve the overall performance of the library.

[V8 article](https://v8.dev/blog/fast-async)

JakeChampion:
PR for this work -- https://github.com/Financial-Times/polyfill-library/pull/398

PR #398, opened 2020-01-24
Refactor code base to use async-await more and other es6+ features

JakeChampion:
Biggest performance win in this is replacing a set of nested loops with a single `.includes` call.

Hopefully this refactoring helps make the code more maintainable as well.

resolves https://github.com/Financial-Times/polyfill-library/issues/177


lib/index.js
```diff
@@ -2,41 +2,41 @@
...
 
 module.exports = {
-	listAliases,
-	listAllPolyfills,
-	describePolyfill,
-	getOptions,
-	getPolyfills,
-	getPolyfillString,
-	normalizeUserAgent
+  listAliases,
```
chee: switching from tabs to spaces in some files?

lib/index.js
```diff
@@ -60,27 +60,33 @@ function describePolyfill(featureName) {
...
-	};
+  const hasProperty = (prop, obj) =>
+    Object.prototype.hasOwnProperty.call(obj, prop);
+  const options = {
+    callback:
+      hasProperty("callback", opts) &&
+      typeof opts.callback === "string" &&
+      opts.callback.match(/^[\w\.]+$/)
+        ? opts.callback
+        : false,
```
chee: standard es6 syntax

lib/index.js
```diff
@@ -99,100 +105,98 @@ function getOptions(opts = {}) {
  * @return {Promise<Object>} - Canonicalised feature definitions filtered for UA
  */
 function getPolyfills(opts) {
```
chee: this is nicer

lib/index.js
JakeChampion: oops! I'll fix that!

PR #406, opened 2020-01-27
do not serve screen.orientation to firefox 30 or above

JakeChampion:
browser support from MDN -- https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation#Browser_compatibility


polyfills/screen/orientation/config.toml
```diff
@@ -11,7 +11,6 @@ edge_mob = "*"
 ie = "9 - *"
 ie_mob = "*"
 firefox = "<30"
-firefox = "*"
```
chee: lol

PR #408, opened 2020-01-27
Update targeting for navigator.sendBeacon

JakeChampion:
Browser support from MDN -- https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon#Browser_compatibility

no longer targeting Edge 14, Safari 11.1, iOS 11.3, Android 5, and above.


polyfills/navigator/sendBeacon/config.toml
```diff
@@ -7,15 +7,15 @@ license = "MIT"
...
-edge_mob = "*"
+edge = "<14"
+edge_mob = "<14"
 firefox = "<31"
 ie = "*"
 opera = "<26"
-safari = "*"
-android = "*"
+safari = "<11.1"
+android = "<5"
```
notlee: "Android webview" on mdn says `40`. Samsung Internet is `4.0`, is this right?

JakeChampion: This is correct, Android prior to 4.4/5 used the stock Android browser which was a fixed version of Chrome, Android 4.4/5 and above us a rolling release of Chromium and as such we parse/detect them as Chrome and not as Android. Whenever you see Android Webview and not Android in MDN you can know that we should not target Android versions 4.4/5 or above.

The reason we target 5 and not 4.4 is because it was mostly rolled out in Android 4.4.4 and polyfill-library does not target patch versions.

I did not update Samsung Internet, good catch, I'll update it now

JakeChampion: Ah, We do not support Samsung Internet versions below 4, that is why I didn't include it. No harm included it though.

PR #410, opened 2020-01-27
Update targeting for Array.prototype.find

JakeChampion:
browser support from MDN -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Browser_compatibility


polyfills/Array/prototype/find/config.toml
```diff
@@ -15,16 +15,16 @@ docs = "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global
...
+edge_mob = "<12"
 firefox = "<24"
 ie = "*"
 ie_mob = "10 - *"
 opera = "<32"
-safari = "* - 7.1"
-ios_saf = "* - 8.0"
-ios_chr = "* - 8.0"
-android = "*"
+safari = "<7.1"
```
notlee:
```suggestion
safari = "<8.0"
```

Same as mdn says and`ios_saf`, et?

JakeChampion: thank you for the spot, I refactored this incorrectly 🙈

PR #412, opened 2020-01-27
update targeting for annex-b String.prototype methods based on MDN browser support tables

JakeChampion:
They all have the nearly same support which is why I have done them together

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/strike#Browser_compatibility


polyfills/String/prototype/strike/config.toml
```diff
@@ -5,22 +5,21 @@ dependencies = [
...
-# opera = "*"
-op_mini = "*"
-# op_mob = "*"
-# safari = "*"
-firefox_mob = "< 4"
-# samsung_mob = "*"
+android = "<1"
+bb = "<1"
+chrome = "<1"
+edge = "<1"
```
notlee: 12?

polyfills/String/prototype/anchor/config.toml
```diff
@@ -5,22 +5,20 @@ dependencies = [
...
-op_mini = "*"
-# op_mob = "*"
-# safari = "*"
-firefox_mob = "< 4"
-# samsung_mob = "*"
-
+android = "<1"
+bb = "<1"
+chrome = "<1"
+edge = "<1"
```
notlee: 12? ... according to mdn... but I feel like these features would have been supported for forever

JakeChampion: 12 is the first version of edge but i'll update anyway

notlee: I did not know this

PR #413, opened 2020-01-27
update String.prototype.repeat targeting based on MDN browser support table

JakeChampion:
mdn browser support table -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat#Browser_compatibility


polyfills/String/prototype/repeat/config.toml
```diff
@@ -9,16 +9,16 @@ docs = "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global
...
 bb = "*"
-chrome = "* - 40"
-firefox = "3.6 - 23"
-opera = "10 - 18"
-safari = "*"
-edge = "*"
-edge_mob = "*"
+chrome = "<41"
+firefox = "<24"
+opera = "<18"
```
notlee: mdn doesn't mention a specific opera version, think this polyfill is needed?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat#Browser_compatibility

JakeChampion: I will keep it in for now, later when we have automation for this, we will know whether it is required :-)

PR #414, opened 2020-01-27
update targeting for ResiveObserver

JakeChampion:
browser support from MDN -- https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#Browser_compatibility


polyfills/ResizeObserver/config.toml
```diff
@@ -7,14 +7,15 @@ license = "MIT"
...
 edge = "*"
 edge_mob = "*"
-firefox = "*"
+firefox = "<69"
 ie = "*"
-opera = "*"
+opera = "<51"
 safari = "*"
-android = "<67"
+android = "<5"
```
notlee: 64?

JakeChampion: Android webview versions are normalised to Chrome. The android targeting in polyfill-library only works for android versions 4.4.3 and below because Android 4.4/5 moved to using a rolling Chromium release for Android webviews. If you see Android Webview mentioned on MDN and not Android, then that translates to a polyfill-library config of android <5

PR #416, opened 2020-01-27
update browser targeting for Object.preventExtensions

JakeChampion:
mdn support table -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions#Browser_compatibility


polyfills/Object/preventExtensions/config.toml
```diff
@@ -6,11 +6,11 @@ spec = "https://tc39.github.io/ecma262/#sec-object.preventextensions"
 docs = "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions"
 
 [browsers]
-android = "*"
+android = "<5"
```
notlee: `<5`?

JakeChampion: 1 does not have the ES2015 definition, no android versions do, only android webview versions which were rolled out in android version 5

PR #435, opened 2020-01-27
update Array.findIndex targeting based on MDN data

JakeChampion:
mdn table -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Browser_compatibility


polyfills/Array/prototype/findIndex/config.toml
```diff
@@ -15,16 +15,16 @@ docs = "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global
 [browsers]
 bb = "6 - 10"
 chrome = "<=45"
```
notlee: Looks like Chrome 45 supports `Array.findIndex`?

```suggestion
 chrome = "<45"
```

PR #438, opened 2020-01-27
update Blob targeting based on MDN data

JakeChampion:
mdn table -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Blob#Browser_compatibility


notlee:
@JakeChampion out of interest are you able to tell in an automated way that this polyfill doesn't cover the `Blob` methods which IE doesn't support, or is it by chance?

PR #442, opened 2020-01-27
update Element.prototype.before targeting based on MDN data

JakeChampion:
mdn table -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ChildNode/before#Browser_compatibility


polyfills/Element/prototype/before/config.toml
```diff
@@ -9,17 +9,17 @@ spec = "http://dom.spec.whatwg.org/#dom-childnode-before"
...
-android = "*"
+android = "<5"
 bb = "*"
 chrome = "<54"
-edge = "*"
-edge_mob = "*"
+edge = "<17"
+edge_mob = "<17"
 firefox = "<49"
 ios_chr = "*"
```
notlee: ` ios_chr = "*" ` be gone (in a future commit)

PR #461, opened 2020-02-10
Add queueMicrotask polyfill

chivchila:


polyfills/queueMicrotask/tests.js
```diff
@@ -0,0 +1,49 @@
...
+
+	it('throws type error if an argument is null', function(){
+		proclaim["throws"](function() { window.queueMicrotask(null) }, TypeError);
+	});
+
+	it('throws type error if an argument is of a String type', function(){
+		proclaim["throws"](function() { window.queueMicrotask('test') }, TypeError);
+	});
+
+	it('array elements are inserted in correct order', async () => {
```
JakeChampion: We can't use async functions or arrow functions, they will be a parse error in es5 and es3 browsers like internet explorer. This will need to be a normal function instead, where await is used, we would need to use return instead and a .then call

chivchila: thank you, Jake!

polyfills/queueMicrotask/polyfill.js
```diff
@@ -0,0 +1,19 @@
+(function (global) {
+	var promise = global.Promise.resolve();
+	global.queueMicrotask = function queueMicrotask(microtask) {
+		if (arguments.length > 1) {
+			throw new TypeError();
+		}
+		if (typeof(microtask) != 'function'){
```
chee:
```suggestion

		if (typeof(microtask) != 'function'){
```

polyfills/queueMicrotask/polyfill.js
```diff
@@ -0,0 +1,20 @@
+(function (global) {
+	var promise = global.Promise.resolve();
+	global.queueMicrotask = function queueMicrotask(microtask) {
+		if (arguments.length > 1) {
+			throw new TypeError();
```
JakeChampion:
```suggestion
			throw new TypeError("Window.queueMicrotask requires at least 1 argument, but only 0 were passed");
```

polyfills/queueMicrotask/polyfill.js
```diff
@@ -0,0 +1,20 @@
...
+		}
+
+		if (typeof(microtask) != 'function'){
+			throw new TypeError();
```
JakeChampion:
```suggestion
			throw new TypeError('Argument 1 of Window.queueMicrotask is not callable.');
```

Issue #338, opened 2019-12-11
nextElementSibling, previousElementSibling polyfills


nickcolley:
Hey folks,

We are using nextElementSibling and previousElementSibiling polyfills that were in the `polyfill-service` library, but they don't seem to be in the new repo.

Is there any documentation why this is? I would be interested in adding these back to this version if you'd be interested too.

https://github.com/Financial-Times/polyfill-service/pull/1062/files#diff-b45a1197b842728cb76b624b6ba7d739

JakeChampion:
Hi, I'm not sure how we lost those polyfills, sorry about that 😔 

We would be interested in getting them added back.

Are you able to submit a pull-request which adds them?

nickcolley:
Yeah, do you have any guidelines or perhaps a pull request that demonstrates a good example of adding something? Otherwise I can figure it out...

nickcolley:
Found this: https://github.com/Financial-Times/polyfill-library/blob/master/.github/contributing.md

Will put together a PR :)

JakeChampion:
Browser support:
https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling#Browser_Compatibility

We would create a new folder inside ./polyfills/Node/prototype named nextElementSibling and create 4 files within this folder:
polyfill.js -- The polyfill code itself
detect.js -- An expression which detects if nextElementSibling exists in the user-agent ( `"nextElementSibling" in Node.prototype` would be a good expression to use)
config.toml -- The configuration which states any dependencies the polyfill has on other features/polyfills and what user-agents/browsers need the polyfill
tests.js -- A test suite to prove the polyfill works correctly

The specification for the feature is here -- https://dom.spec.whatwg.org/#dom-nondocumenttypechildnode-nextelementsibling

Some tests already exist for this feature within the web-platform-tests projects, we should/could base our tests off of those ones. https://github.com/web-platform-tests/wpt/blob/master/dom/nodes/Element-nextElementSibling.html

A good folder in this project to look at if unsure about what to do would be requestIdleCallback https://github.com/Financial-Times/polyfill-library/tree/master/polyfills/requestIdleCallback which @notlee contributed.

nickcolley:
Thanks <3

JakeChampion:
If you have any questions, please comment on this issue and I will do my best to help :-)

nickcolley:
[...]

PR #484, opened 2020-02-24
Add nextElementSibling and previousElementSibling polyfills

nickcolley:
Raising this to see if I'm on the right track.

## Do we need to polyfill CharacterData?
So it looks like this feature is used in the node type of `NonDocumentTypeChildNode` since it can be `Element` and `CharacterData`. Does this block this improvement, or can we continue and consider a wider `NonDocumentTypeChildNode` polyfill later that would also sort CharacterData?

- [NonDocumentTypeChildNode.nextElementSibling
](https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling#Polyfill_for_Internet_Explorer_9_and_Safari)

Closes https://github.com/Financial-Times/polyfill-library/issues/338

### DIFF
--- a/polyfills/Element/prototype/nextElementSibling/config.toml
+++ b/polyfills/Element/prototype/nextElementSibling/config.toml
@@ -1,0 +1,25 @@
+aliases = [ ]
+dependencies = [
+  "Element",
+  "Object.defineProperty"
+]
+license = "MIT"
+docs = "https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling"
+
+[browsers]
+# android = "all versions natively support this feature"
+# bb = "Not on MDN"
+chrome = "<4"
+edge = "<12"
+# edge_mob = "Not on MDN"
+firefox = "<3.5"
+# ios_chr = "Not on MDN"
+# ios_saf = "all versions natively support this feature"
+ie = "<9"
+# ie_mob = "Not on MDN"
+opera = "<10"
+# op_mini = "Not on MDN"
+op_mob = "<10.1"
+safari = "<4"
+firefox_mob = "<4"
+# samsung_mob = "all versions natively support this feature"
--- a/polyfills/Element/prototype/nextElementSibling/detect.js
+++ b/polyfills/Element/prototype/nextElementSibling/detect.js
@@ -1,0 +1,1 @@
+'document' in self && "nextElementSibling" in document.documentElement
--- a/polyfills/Element/prototype/nextElementSibling/polyfill.js
+++ b/polyfills/Element/prototype/nextElementSibling/polyfill.js
@@ -1,0 +1,7 @@
+Object.defineProperty(Element.prototype, "nextElementSibling", {
+	get: function(){
+		var el = this.nextSibling;
+		while (el && el.nodeType !== 1) { el = el.nextSibling; }
+		return el;
+	}
+});
--- a/polyfills/Element/prototype/nextElementSibling/tests.js
+++ b/polyfills/Element/prototype/nextElementSibling/tests.js
@@ -1,0 +1,45 @@
+it("should return null if the node is the only child of its parent node", function () {
+  var parent = document.createElement('div'),
+      p = document.createElement('p');
+  parent.appendChild(p);
+
+  proclaim.strictEqual(p.nextElementSibling, null);
+});
+
+it("should return null if the node only has text sibling", function () {
+  var parent = document.createElement('div'),
+      p = document.createElement('p');
+      text = document.createTextNode('Hi there, how are you doing today?');
+  parent.appendChild(p);
+  parent.appendChild(text);
+
+  proclaim.strictEqual(p.nextElementSibling, null);
+});
+
+it("should return null if the node only has comment sibling", function () {
+  var parent = document.createElement('div'),
+      p = document.createElement('p');
+      comment = document.createComment('This is a comment in the document.');
+  parent.appendChild(p);
+  parent.appendChild(comment);
+
+  proclaim.strictEqual(p.nextElementSibling, null);
+});
+
+it("should return the first child element", function () {
+  var parent = document.createElement('div'),
+      h2 = document.createElement('h2'),
+      p1 = document.createElement('p'),
+      p2 = document.createElement('p'),
+      p3 = document.createElement('p'),
+      text = document.createTextNode('Hi there, how are you doing today?'),
+      comment = document.createComment('This is a comment in the document.');
+  parent.appendChild(h2);
+  parent.appendChild(text);
+  parent.appendChild(comment);
+  parent.appendChild(p1);
+  parent.appendChild(p2);
+  parent.appendChild(p3);
+
+  proclaim.strictEqual(h2.nextElementSibling, p1);
+});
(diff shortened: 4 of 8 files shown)


polyfills/Element/prototype/nextElementSibling/detect.js
```diff
@@ -0,0 +1 @@
+'document' in self && "nextElementSibling" in document.documentElement
```
nickcolley: Copied `'document' in self` from tests next to this one for consistency but not sure it's necessary.

polyfills/Element/prototype/nextElementSibling/config.toml
```diff
@@ -0,0 +1,25 @@
+aliases = [ ]
+dependencies = [
+  "Element",
+  "Object.defineProperty"
+]
+license = "MIT"
+docs = "https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling"
+
+[browsers]
```
nickcolley: Do you have any documentation on how to populate this, it's not clear to me what syntax it required and what this is used for. Finally MDN does not have all the data required to fill this in.

polyfills/Element/prototype/nextElementSibling/config.toml
```diff
@@ -0,0 +1,25 @@
+aliases = [ ]
```
nickcolley: Not sure it there's any aliases required, do you have any guidance on the schema for this file generally?

polyfills/Element/prototype/nextElementSibling/tests.js
```diff
@@ -0,0 +1,45 @@
+it("should return null if the node is the only child of its parent node", function () {
```
nickcolley: Tests are from the original work, happy to improve or change these.

https://github.com/Financial-Times/polyfill-service/pull/1062/files#diff-b45a1197b842728cb76b624b6ba7d739

polyfills/Element/prototype/nextElementSibling/polyfill.js
```diff
@@ -0,0 +1,7 @@
+Object.defineProperty(Element.prototype, "nextElementSibling", {
```
nickcolley: In production on GOV.UK: https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/vendor/polyfills/Element/prototype/nextElementSibling.js

polyfills/Element/prototype/previousElementSibling/polyfill.js
```diff
@@ -0,0 +1,7 @@
+Object.defineProperty(Element.prototype, 'previousElementSibling', {
```
nickcolley: In production on GOV.UK: https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/vendor/polyfills/Element/prototype/previousElementSibling.js

polyfills/Element/prototype/nextElementSibling/config.toml
JakeChampion: The schema for this file hasn't been written yet, sorry 😔 

We don't need any alias for this polyfill. An alias is another name for the polyfill that can be used when requesting a bundle, E.G. We have an alias `es5` for all the polyfills to replicate an ECMAScript 5 environment

polyfills/Element/prototype/nextElementSibling/config.toml
JakeChampion: This section is a list of browser-name to semver range strings. The semver range string should match all versions which need the polyfill. I.E. Internet Explorer 9 added support for Element.prototype.nextElementSibling so the semver range would be `<9` to match versions below 9.

polyfills/Element/prototype/nextElementSibling/config.toml
```diff
@@ -0,0 +1,25 @@
...
+dependencies = [
+  "Element",
+  "Object.defineProperty"
+]
+license = "MIT"
+docs = "https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling"
+
+[browsers]
+android = "todo"
```
JakeChampion:
```suggestion
# android = "all android version natively support this feature"
```

polyfills/Element/prototype/nextElementSibling/config.toml
```diff
@@ -0,0 +1,25 @@
...
+bb = "todo"
```
JakeChampion: For browsers which MDN does not have, we should comment them out. At a later date we can manually test on those browsers and see which versions require the polyfill
```suggestion
# bb = "Not on MDN"
```

polyfills/Element/prototype/nextElementSibling/polyfill.js
JakeChampion: 🙌

polyfills/Element/prototype/nextElementSibling/tests.js
nickcolley: Think I'll need to adjust these regardless, since this polyfill doesnt return null anymore.
Might be good to have a case for when you check the nextElementSibling on the last child too.

nickcolley: I was having trouble running the tests:

`npm run test-polyfills -- --features=Element.prototype.nextElementSibiling`

It passes with 0 tests run, need to dig into it...

JakeChampion: The typo in sibling is likely why

nickcolley: 😂 that'll be it, thanks :)

polyfills/Element/prototype/nextElementSibling/polyfill.js
```diff
@@ -0,0 +1,7 @@
...
+	get: function(){
+		var el = this.nextSibling;
+		while (el && el.nodeType !== 1) { el = el.nextSibling; }
+		return el;
```
nickcolley: The spec says return null if no element so I need to adjust this as I think it'll return undefined or something else falsey instead of explicitly null

> and null otherwise.

nickcolley: Hmm testing in Internet Explorer 8, it is null already, but perhaps we should add a check in to be super sure. Will see what reviewers think.

nickcolley:
@JakeChampion think this is ready for another review, I believe the tests have not run since I'm an external contributor and don't have access to secure credentials?

I've also left a question about `CharacterData` in the main pull request description I'm not sure about.

polyfills/Element/prototype/nextElementSibling/polyfill.js
JakeChampion: this.nextSibling is null if there is no next sibling, this code looks to be spec compliant 👍

JakeChampion:
Tests are now running on CircleCI :-) Thank you for the pull request!

JakeChampion:
> ## Do we need to polyfill CharacterData?
> 
> So it looks like this feature is used in the node type of `NonDocumentTypeChildNode` since it can be `Element` and `CharacterData`. Does this block this improvement, or can we continue and consider a wider `NonDocumentTypeChildNode` polyfill later that would also sort CharacterData?

A wider `NonDocumentTypeChildNode` can be added later 👍

nickcolley:
Thanks for sorting those issues for me :+1:

JakeChampion:
@nickcolley Thank you so much for contributing these polyfills! They will be in the next release :-)

PR #485, opened 2020-02-24
add Node.prototype.getRootNode

JakeChampion:


chee:
does it need to do something like this https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Node/prototype/contains/polyfill.js for IE? (i.e. trying HtmlElement or Element if there's no `Node`)

Issue #498, opened 2020-02-25
Documentation on how to add a polyfill


nickcolley:
Type: feature request

## What
Adding documentation for contributors to the project so they know what's needed to add a new polyfill

## Details
I had a lot of very kind help from @JakeChampion but wondering if you'd welcome a pull request that documents some of the advice for example how the toml schema works and [other things mentioned in our discussion](https://github.com/Financial-Times/polyfill-library/issues/338).

You're welcome to close this issue depending on what you decide.

JakeChampion:
That would be very much appreciated :-)

PR #501, opened 2020-02-25
Add documentation explaining how to add a polyfill

nickcolley:
This ended up fairly lengthy although I tried to keep it to only the necessary information.

Perhaps this would work better in a different document rather than the main contributing document.

As always, I welcome any feedback and if you decide you don't want to take this proposal at all that is okay.

[See the changes rendered](https://github.com/Financial-Times/polyfill-library/blob/f6de8a622d3cd457c6dbed75796792ce0171078e/.github/contributing.md#adding-a-new-polyfill) 

Closes https://github.com/Financial-Times/polyfill-library/issues/498

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -11,6 +11,127 @@ the developers managing and developing this open source project. In return,
 they should reciprocate that respect in addressing your issue or assessing
 patches and features.
 
+## Adding a new polyfill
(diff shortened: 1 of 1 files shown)


JakeChampion:
I think we should look to add this to the website as well :-)

PR #510, opened 2020-02-27
always run smoothscroll polyfill - fix #508

### LINKED ISSUE #508: [bug] smoothscroll polyfill is never executed
## What
The whole smoothscroll polyfill is wrapped in this:
```js
if (!(!0)) {
 /* polyfill */
}
```
Which is boolean for `if(false) {}`. Therefore it's never run.

## Details
I'm not familiar with this repo, but I guess commit number https://github.com/Financial-Times/polyfill-library/commit/dc90a49d958bc8f8b4baa70297aeffcdc539edca is the culprit.
I also guess that the content of the polyfill [detect.js](https://github.com/Financial-Times/polyfill-library/blob/dc90a49d958bc8f8b4baa70297aeffcdc539edca/polyfills/smoothscroll/detect.js) should be `false` and not `true` (I've made a little PR in case this is correct)

rasteiner:
the polyfill does feature detection inside

### DIFF
--- a/polyfills/smoothscroll/detect.js
+++ b/polyfills/smoothscroll/detect.js
@@ -2,4 +2,4 @@
 // "scrollBy" in Element.prototype &&
 // "scroll" in Element.prototype &&
 // "scrollTo" in Element.prototype
-true
+false


JakeChampion:
We've added a feature detect which can detect polyfills as well -- https://github.com/Financial-Times/polyfill-library/pull/515

Issue #109, opened 2019-02-06
Add: {DateTimeFormat,Number}.prototype.formatToParts polyfills


JakeChampion:
These are stage 4 in the proposal process -- https://github.com/tc39/proposal-intl-formatToParts

RichiCoder1:
Is there a specific way we could help move this forward? Would love to help, especially since this makes using libraries likes luxon (https://moment.github.io/luxon/docs/manual/matrix.html#intl-tokens) much more straightforward.

JakeChampion:
Hi @RichiCoder1, if you want to add this feature to polyfill-library then I would be very happy to review your work and guide you through the process of [contributing](https://github.com/Financial-Times/polyfill-library#contributing).

I think it might be simpler to have these as their own polyfills rather than monkeypatching but I am happy to accept a pull-request which does monkeypatching if that is what you prefer :-)

longlho:
This cannot be monkey-patched (we tried). I suggest adding https://www.npmjs.com/package/@formatjs/intl-unified-numberformat polyfill which has `formatToParts` along with all stage-3 features in unified `Intl.NumberFormat`

longlho:
Can I get a release please:)

JakeChampion:
v3.95.0 is now released 🎉

PR #521, opened 2020-03-02
feat: add base version of DateTimeFormat.formatToParts polyfill

Tasyp:
Addresses issue #109 

I am proposing to use a temporary solution for polyfilling the `DateTimeFormat.protope.formatToParts` method.

It has a few drawbacks:
   1. It will make the bundle size bigger than it could be (e.g. separate polyfill for one method)
   2. The bigger amount of browsers will receive the `Intl` polyfill

There will be no alternatives until [#242](https://github.com/andyearnshaw/Intl.js/issues/242) is solved and I don't think it is viable to maintain a custom polyfill. It is better to focus on improving the `Intl.js` original polyfill rather than losing focus and rewriting the same code again.


JakeChampion:
Thanks for the PR @Tasyp 
It looks to not pass the tests when running in Internet Explorer 8, would you be able to look into making the tests pass in that browser at all?
```
Failures:
 - ie/8.0:
    -> Intl.DateTimeFormat.prototype.formatToParts passes the feature detect
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Intl.DateTimeFormat.prototype.formatToParts
       false == true
```

PR #528, opened 2020-03-04
update targets for polyfills/console/timelineEnd/config.toml

JakeChampion:


notlee:
These don't match with https://developer.mozilla.org/en-US/docs/Web/API/Console/timeEnd ?

JakeChampion:
@notlee Correct, that is because this is `timelineEnd` and not `timeEnd`. `timelineEnd` was removed from browsers before console was standardised. https://bugzilla.mozilla.org/show_bug.cgi?id=1351795

PR #531, opened 2020-03-05
Ship Math.hypot to chrome 77

JakeChampion:
as it's implementation does not return `NaN` for `Math.hypot(undefined,0)` and instead returns `0`


notlee:
Curious! So 38-77 does return `NaN`?

JakeChampion:
@notlee Correct, it was a regression they had in only 77

PR #539, opened 2020-03-12
abortcontroller : update config.toml

romainmenke:
The `config.toml` seems to have old info from the original pull request where another polyfill was considered.

This change updates the config to reflect the current polyfill.


JakeChampion:
Thank you for fixing this @romainmenke

PR #558, opened 2020-04-04
 Refactor: inline the polyfill resolver into a non-recursive for loop 

JakeChampion:


lib/sources.js
```diff
@@ -19,16 +19,12 @@ const polyfillDirectory = path.join(__dirname, "../polyfills/__dist");
...
   if (meta === undefined) {
-    try {
-      meta = await readFile(
-        path.join(polyfillDirectory, featureName, "meta.json"),
-        "UTF-8"
-      );
-      meta = JSON.parse(meta);
-      polyfillMetaCache.set(featureName, meta);
-    } catch (error) {
-      // This is naughty.
```
JakeChampion, line 30: I didn't want us to carry on being naughty

polyfills/Symbol/prototype/description/config.toml
```diff
@@ -2,7 +2,8 @@ aliases = [ ]
 dependencies = [
     "Symbol",
     "Object.defineProperty",
-    "String.prototype.trim"
+    "String.prototype.trim",
+    "_ESAbstract.Type"
```
JakeChampion, line 6: This was missing and the old polyfill bundling logic has a bug where it added this dependency! We had a bug fixing another bug 🙈

chee: amazing. unexpected

lib/index.js
```diff
@@ -104,99 +112,97 @@ function getOptions(options_ = {}) {
...
-    const meta = await sources.getPolyfillMeta(featureName);
-    const result = ((meta && meta.dependencies) || [])
-      .filter(depName => !options.excludes.includes(depName))
-    .concat(featureName);
-    return result;
-  };
-  const resolveDependencies = createAliasResolver(aliasDependencies);
+  let ua = getNormalisedUA(options.uaString);
+  const features = new Set(Object.keys(options.features));
+  const targetedFeatures = {};
```
chee, line 119: what's the difference between features and targetedFeatures

lib/index.js
```diff
@@ -104,99 +112,97 @@ function getOptions(options_ = {}) {
...
-      meta.browsers &&
-      meta.browsers[ua.getFamily()] &&
-      ua.satisfies(meta.browsers[ua.getFamily()]);
-    const hasAlwaysFlagOverride = features[featureName].flags.has("always");
-        const unknownOverride = options.unknown === "polyfill" && ua.isUnknown();
-    
-        if (unknownOverride) {
-          features[featureName].flags.add("gated");
-        }
+  function removeFeature(featureName) {
```
chee, line 121: maybe we could have a comment here explaining why a feature would be removed

lib/index.js
chee: maybe `features` should be `featureNames`?

chee:
you are removing special alias logic to instead make the "alias" a virtual dependent. is that right?

lib/index.js
JakeChampion: A feature would be removed in these scenarios:
- If it has been specified to be excluded from the bundle via the `exlcudes` option.
- If the feature is not meant to be in the bundle it does not target the browser. Targeting a browser happens if the feature has the `always` option applied or the browser is unsupported/unknown and the `unknown` option is set to `polyfill` or the browser version is within the semver range specified in the features config.toml file.

lib/index.js
JakeChampion: That's a better name for it :-)

JakeChampion:
> you are removing special alias logic to instead make the "alias" a virtual dependent. is that right?

I'm hoping to have the code be more understandable and maintainable. The name of the old function `createAliasResolver` was an incorrect name, since it did much more than resolve aliases. It was the abstracted logic for figuring out what features/polyfills to include in a bundle.

lib/index.js
```diff
@@ -104,99 +112,97 @@ function getOptions(options_ = {}) {
...
```
JakeChampion, line 121:
```suggestion
  /*
    A feature would be removed in these scenarios:
    - If it has been specified to be excluded from the bundle via the exlcudes option.
    - If the feature is not meant to be in the bundle it does not target the browser. 
        Targeting a browser happens if the feature has the always option applied or 
        the browser is unsupported/unknown and the unknown option is set to polyfill 
        or the browser version is within the semver range specified in the features config.toml file.
  */
  function removeFeature(featureName) {
```

PR #569, opened 2020-04-10
Add CSS.supports polyfill

### LINKED ISSUE #563: CSS.supports polyfill is missing
It would be nice to have `CSS.supports` polyfill, yesterday our production went down due to this polyfill :) Here is the souce code: https://github.com/termi/CSS.supports

MoritzLoewenstein:
fixes #563 This polyfill has 3 connected minor issues:
1. It doesnt support the parentheses-less one argument version  
2. On presto-based opera browsers (12.1, op_mini) it just uses window.supportsCSS with whatever limitations this has (i couldnt find out, maybe the same as 1.).  
3. If there is an earlier spec of window.CSS.supports supported, it will be used since the polyfill doesnt exceed the functionality of the earlier implementation (this affects Edge 12-18, Ff 22-54 & Chrome 28-60).

I dont really know how reliable this method of detecting CSS compatiblility is, but I did not run into issues.


MoritzLoewenstein:
Is there generally something wrong with my pr?
If I can do anything to speed this up, let me know.

Issue #211, opened 2019-05-23
Feature Request: ResizeObserver polyfill


andrewgadziksonos:
## What
Would like to add the [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) polyfill

## Details

The `ResizeObserver` interface reports changes to the content rectangle of an Element or the bounding box of an SVGElement. The content rectangle is the box in which content can be placed, meaning the border box minus the padding. 

Support is rapidly gaining momentum, so it would be nice to manage this polyfill in this service.  
https://caniuse.com/#feat=resizeobserver

PR submitted https://github.com/Financial-Times/polyfill-library/pull/212

aronwoost:
Can be closed, because PR was merged. 👍

nicooprat:
Looks like the repo is abandoned, and I had a lot of issues with this polyfill especially on Safari 13 and IE 11. This one looks more robust and updated: https://github.com/juggle/resize-observer/ Would you mind use this one instead? 🙏

JakeChampion:
Sure thing @nicooprat, are you able to contribute that change to the project?

nicooprat:
@JakeChampion Here's my first try ;) https://github.com/Financial-Times/polyfill-library/pull/644

PR #644, opened 2020-05-21
Replace ResizeObserver polyfill by a more robust and maintained one

nicooprat:
I had some troubles using the old ResizeObserver polyfill, especially with Safari 13 and IE 11. I tried this one which solved all my issues and seems way more maintained.

* Old repo: https://github.com/que-etc/resize-observer-polyfill (last commit end of 2018)
* New repo: https://github.com/juggle/resize-observer/ (last commit < 1 month ago)

Please tell me if I'm missing something in this PR.

* The command `npm run build` works fine. 
* I don't know if I'm supposed to add tests? The old polyfill didn't have any in this repo, and the new one has its own in its repo.
* The `detect.js` script has no reason to change.

Ref: https://github.com/Financial-Times/polyfill-library/issues/211#issuecomment-630720483


JakeChampion:
Thank you for the PR @nicooprat, it is greatly appreicated :-)
I added some basic tests for the polyfill (the basic tests from the original repo) and updated the polyfill to export the globals correctly, it looked to be exporting them as `global.ResizeObserver.ResizeObserver` and `global.ResizeObserver.ResizeObserverEntry` when they should be `global.ResizeObserver` and `global.ResizeObserverEntry`.

Hopefully the tests pass and we can merge this :-)

nicooprat:
Thanks for your help! If the tests fail, maybe we should check it with the polyfill maintainers?

JakeChampion:
Currently failing on this test:
```
Failures:
 - safari/9.1:
    -> ResizeObserver ResizeObserver Observer should allow trying to unobserve multiple times.
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=ResizeObserver
       done() called multiple times
```

TremayneChrist:
Hi @JakeChampion I maintain the RO polyfill.

Changes have been made recently to switch from using `Map` to `WeakMap`. Can you update the dependencies to reflect this change?

I've tested with the latest version on Safari 9.1 and do not see the issue with the callback being fired multiple times.

JakeChampion:
Hi, I updated the polyfill and added WeakMap as a dependency, some different tests now look to no longer pass.

```
Failures:
- chrome/29.0 upto and including chrome/63.0:
   -> ResizeObserver ResizeObserver Observer should not observe after a disconnect.
      http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=ResizeObserver
      Uncaught AssertionError: false deepStrictEqual true (http://bs-local.com:9876/proclaim.js:402)
- chrome/30.0 and chrome/36.0 and chrome/41.0 and chrome/42.0:
   -> ResizeObserver ResizeObserver "after each" hook for "Observer should return itself in the callback."
      http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=ResizeObserver
      Cannot read property 'removeChild' of null
```

TremayneChrist:
> ResizeObserver Observer should not observe after a disconnect.

This test needs to be updated, as it's actually incorrect and the [spec](https://drafts.csswg.org/resize-observer-1/#dom-resizeobserver-disconnect) says that it should only clear the observations and active list. This has recently been fixed (https://github.com/juggle/resize-observer/issues/104).

The other failure seems to be random and is happening in the "after each" hook for removing child. Could this be to do with the other failing test?

polyfills/ResizeObserver/tests.js
```diff
@@ -359,6 +359,15 @@ describe("ResizeObserver", function() {
 		delay(done)
 	})
 
+	it("Observer should not observe after a disconnect.", function(done) {
```
TremayneChrist, line 362: Observer should ~~not~~ observe after a disconnect.

TremayneChrist:
Hey @JakeChampion is it looking promising, or, quite the opposite? 🙃

JakeChampion:
@TremayneChrist Promising! :-D

TremayneChrist:
@JakeChampion how do you handle updates to dependencies? There is a future update which could use `matchMedia` API and would need a change in polyfill.io to request the additional polyfill for IE9.

PR #659, opened 2020-05-29
Use seamless-scroll-polyfill for smoothscroll

### LINKED ISSUE #657: Replace smoothscroll with seamless-scroll-polyfill
Possibly a bug?

## What
The current `smoothscroll` library, [smoothscroll-polyfill](https://www.npmjs.com/package/smoothscroll-polyfill), doesn't support the `block` and `inline` options. iamdustan/smoothscroll#87 suggests that it's no longer maintained. [seamless-scroll-polyfill](https://www.npmjs.com/package/seamless-scroll-polyfill) is a better-maintained fork that complies with more of the spec.

jamesarosen:
Previously, polyfill-library used smoothscroll-polyfill. That library is no longer well-maintained and is missing some parts of the smoothscroll spec.

This commit swaps in seamless-scroll-polyfill, which is a better-maintained fork of smoothscroll-polyfill.

Resolves #657

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "picturefill": "^3.0.1",
     "resize-observer-polyfill": "^1.5.1",
     "rimraf": "^3.0.0",
-    "smoothscroll-polyfill": "^0.4.4",
+    "seamless-scroll-polyfill": "^1.0.10",
     "spdx-licenses": "^1.0.0",
     "stream-cache": "^0.0.2",
     "stream-from-promise": "^1.0.0",
--- a/polyfills/smoothscroll/patch.jsdiff
+++ b/polyfills/smoothscroll/patch.jsdiff
@@ -1,19 +1,34 @@
 ===================================================================
---- a/polyfills/smoothscroll/polyfill.js
-+++ b/polyfills/smoothscroll/polyfill.js
-@@ -421,13 +421,8 @@
-       }
+--- a/polyfills/smoothscroll/polyfill-original.js
++++ b/polyfills/smoothscroll/polyfill-original.js
+@@ -1,8 +1,5 @@
+-(function (global, factory) {
+-    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
+-    typeof define === 'function' && define.amd ? define(['exports'], factory) :
+-    (global = global || self, factory(global.seamless = {}));
+-}(this, (function (exports) { 'use strict';
++(function () {
++    'use strict';
+
+     /*! *****************************************************************************
+     Copyright (c) Microsoft Corporation. All rights reserved.
+@@ -701,17 +698,7 @@
+         polyfill$2(options);
      };
-   }
- 
--  if (typeof exports === 'object' && typeof module !== 'undefined') {
--    // commonjs
--    module.exports = { polyfill: polyfill };
--  } else {
--    // global
--    polyfill();
--  }
-+  // global
-+  polyfill();
- 
- }());
+
+-    exports.elementScroll = elementScroll;
+-    exports.elementScrollBy = elementScrollBy;
+-    exports.elementScrollIntoView = elementScrollIntoView;
+-    exports.elementScrollTo = elementScroll;
+-    exports.polyfill = polyfill$7;
+-    exports.seamless = polyfill$7;
+-    exports.windowScroll = windowScroll;
+-    exports.windowScrollBy = windowScrollBy;
+-    exports.windowScrollTo = windowScroll;
++    polyfill$7();
+
+-    Object.defineProperty(exports, '__esModule', { value: true });
+-
+-})));
++})();
+//# sourceMappingURL=seamless.js.map
--- a/tasks/node/updatesources.js
+++ b/tasks/node/updatesources.js
@@ -61,7 +61,7 @@ glob('polyfills/**/config.toml', globOptions)
 				}
 			})
             .filter(config => 'install' in config)
-            .forEach(installPolyfill);
+            .forEach(toml => installPolyfill(toml));
         
     })
     .then(() => console.log('Polyfills updated successfully'))
--- a/package-lock.json
+++ b/package-lock.json
@@ -5116,6 +5116,11 @@
       "integrity": "sha512-YZo3K82SD7Riyi0E1EQPojLz7kpepnSQI9IyPbHHg1XXXevb5dJI7tpyN2ADxGcQbHG7vcyRHk0cbwqcQriUtg==",
       "dev": true
     },
+    "seamless-scroll-polyfill": {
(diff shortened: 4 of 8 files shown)


polyfills/smoothscroll/patch.jsdiff
```diff
@@ -1,19 +1,21 @@
...
+
+     /*! *****************************************************************************
+     Copyright (c) Microsoft Corporation. All rights reserved.
+@@ -713,5 +710,5 @@
+
+     Object.defineProperty(exports, '__esModule', { value: true });
+
+-})));
++}(self));
+ //# sourceMappingURL=seamless.js.map
```
jamesarosen, line 21: This isn't quite right. It leaves the following at the bottom of the IIFE, but doesn't define `exports`:
```js
    exports.elementScroll = elementScroll;
    exports.elementScrollBy = elementScrollBy;
    exports.elementScrollIntoView = elementScrollIntoView;
    exports.elementScrollTo = elementScroll;
    exports.polyfill = polyfill$7;
    exports.seamless = polyfill$7;
    exports.windowScroll = windowScroll;
    exports.windowScrollBy = windowScrollBy;
    exports.windowScrollTo = windowScroll;

    Object.defineProperty(exports, '__esModule', { value: true });
```

package.json
```diff
@@ -92,7 +92,7 @@
     "cli-color": "^2.0.0",
     "dotenv": "^8.2.0",
     "eslint": "^6.8.0",
-    "eslint-plugin-unicorn": "^17.2.0",
+    "eslint-plugin-unicorn": "^19.0.1",
```
jamesarosen, line 95: This was from dependabot. I added the corresponding eslint fixes to this PR in a separate commit.

polyfills/smoothscroll/patch.jsdiff
jamesarosen: Good call. Updated.

tasks/node/updatesources.js
```diff
@@ -61,7 +61,7 @@ glob('polyfills/**/config.toml', globOptions)
 				}
 			})
             .filter(config => 'install' in config)
-            .forEach(installPolyfill);
+            .forEach(toml => installPolyfill(toml));
```
jamesarosen, line 64: This and the other changes below are to make `npm run lint` pass now that #655 is merged.

polyfills/smoothscroll/patch.jsdiff
```diff
@@ -1,19 +1,41 @@
...
+-    exports.elementScroll = elementScroll;
+-    exports.elementScrollBy = elementScrollBy;
+-    exports.elementScrollIntoView = elementScrollIntoView;
+-    exports.elementScrollTo = elementScroll;
+-    exports.polyfill = polyfill$7;
+-    exports.seamless = polyfill$7;
+-    exports.windowScroll = windowScroll;
+-    exports.windowScrollBy = windowScrollBy;
+-    exports.windowScrollTo = windowScroll;
++    polyfill$7();
```
jamesarosen, line 28: This runs
```js
var polyfill$7 = function (options) {
    if ("scrollBehavior" in document.documentElement.style) {
        return;
    }
    polyfill$4(options);
    polyfill$6(options);
    polyfill$5(options);
    polyfill(options);
    polyfill$3(options);
    polyfill$1(options);
    polyfill$2(options);
};
```
I won't list out all the parts, but for example
```js
var polyfill$4 = function (options) {
    var originalFunc = getOriginalFunc$2();
    window.scroll = function scroll() {
        var _a = __read(arguments, 2), _b = _a[0], arg0 = _b === void 0 ? 0 : _b, _c = _a[1], arg1 = _c === void 0 ? 0 : _c;
        if (typeof arg0 === "number" && typeof arg1 === "number") {
            return originalFunc.call(this, arg0, arg1);
        }
        if (Object(arg0) !== arg0) {
            throw new TypeError("Failed to execute 'scroll' on 'Window': parameter 1 ('options') is not an object.");
        }
        return windowScroll(__assign(__assign({}, arg0), options));
    };
};
```

polyfills/smoothscroll/patch.jsdiff
```diff
@@ -1,19 +1,41 @@
...
++    self.elementScroll = elementScroll;
++    self.elementScrollBy = elementScrollBy;
++    self.elementScrollIntoView = elementScrollIntoView;
++    self.elementScrollTo = elementScroll;
++    self.windowScroll = windowScroll;
++    self.windowScrollBy = windowScrollBy;
++    self.windowScrollTo = windowScroll;
```
jamesarosen, line 35: I'm actually not sure these are needed. They're not part of the spec or in the MDN docs. I don't see them in my browser either. I think they're just functions that the module exports so you can use them without patching `window` et al.

jamesarosen:
Is there something I need to do to get the rest of the checks to pass? The tests are passing.

JakeChampion:
@jamesarosen I've triggered CI to run once more :-)

JakeChampion:
it looks to pass the tests but not the feature detect in detect.js 
```
Failures:
 - chrome/29.0 up to and including chrome/60.0:
    -> smoothscroll passes the feature detect
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=smoothscroll
       false == true
```

jamesarosen:
Does
```
Failures:
 - chrome/29.0 up to and including chrome/60.0:
    -> smoothscroll passes the feature detect
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=smoothscroll
       false == true
```
mean that it expects the polyfill to be required but `detect.js` says it's not? Or that after applying the `smoothscroll` polyfill, it should _now_ pass `detect.js`?

I didn't change `detect.js`. In what way might changing the polyfill implementation cause those tests to fail?

JakeChampion:
It means that after applying the smoothscroll polyfill, the `detect.js` expression should be evaluate to `true` but it did not.

jamesarosen:
When given an object (as `detect.js` does) their implementation of [`Element.prototype.scrollTo`](https://github.com/magic-akari/seamless-scroll-polyfill/blob/master/src/Element.scrollTo.ts) calls `elementScroll`, which [calls](https://github.com/magic-akari/seamless-scroll-polyfill/blob/031fb5d8149ee69fb1706fffeb510126e4a74908/src/Element.scroll.ts#L30) `options.behavior`. The problem is that [`Element.prototype.scrollTo`](https://github.com/magic-akari/seamless-scroll-polyfill/blob/031fb5d8149ee69fb1706fffeb510126e4a74908/src/Element.scrollTo.ts#L20) uses destructuring to merge some objects, which removes the `behavior` property defined via `Object.defineProperty`.

I think we just need to add `enumerable: true` to the property descriptor.

Issue #207, opened 2019-05-19
Feature Request: add polyfill for TextEncoder and TextDecoder


lpd-au:
## What
Add a polyfill for [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) and [TextDecoder](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder) capable of encoding and decoding UTF-8.

## Details
IE11 and EdgeHTML do not have support for the `TextEncoder.prototype.encode` and `TextDecoder.prototype.decode` apis. Native implementations of TextEncoder must only encode to UTF-8 but TextDecoder may support other encodings, however it is imo rare to use any other in practice given it's not reversible, so it would be sufficient to throw a RangeError when constructing a TextDecoder for any other encoding.

My use case for this polyfill is in combination with websockets. Websockets support a text and binary mode, and I would like to use this api to decode strings transmitted within the binary data (when `aWebSocket.binaryType` is set to `arraybuffer`), as well as send encoded strings back to the server.

## Additional Information

MDN has a [working polyfill for TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder#Polyfill) but not for TextDecoder. I submitted a request to that repo but was directed here instead: https://github.com/mdn/sprints/issues/1508.

JakeChampion:
This was added in [`#672`](https://github.com/Financial-Times/polyfill-library/pull/672) and released in [`v3.96.0`](https://github.com/Financial-Times/polyfill-library/releases/tag/v3.96.0)

PR #672, opened 2020-06-18
Add TextEncoder#encode to polyfills

marshallycheng:
#207

This is a preliminary PR for adding `TextEncoder#encode` to the library. Comments and suggestions would be much appreciated.

Some questions I had:
- In `polyfill.js` we use [`Object.isPrototypeOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isPrototypeOf). Does that need to be added to the dependencies list? MDN mentions that it's not supported in IE < 9.
- The constructor function for `TextEncoder` is implemented in the same version as `TextEncoder#encode`, but received some updates to prevent errors from being thrown in later versions of some browsers. Those later versions are listed as the  browser compatibility version for the constructor on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder#Specifications). Does that necessitate any changes to the browser list in `config.toml`? I currently have the earlier versions listed. 
- Noticed that some packages are listed as third-party packages rather than being checked into the source code. This polyfill code comes directly from [FastestSmallestTextEncoder](https://github.com/anonyco/FastestSmallestTextEncoderDecoder). Is there a preference here between the two options?
- I'm not sure how the folder structure for TextEncoder should work. Should `#encode` be in `polyfills/TextEncoder/encode`? The `TextEncoder` class itself is also being polyfilled here so I figured it was okay to just leave everything in `polyfills/TextEncoder`.


JakeChampion:
> In `polyfill.js` we use [`Object.isPrototypeOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isPrototypeOf). Does that need to be added to the dependencies list? MDN mentions that it's not supported in IE < 9.

Yes, we support IE 8, you will need to add `Object.isPrototypeOf` as a dependency in `config.toml` 👍 



> The constructor function for `TextEncoder` is implemented in the same version as `TextEncoder#encode`, but received some updates to prevent errors from being thrown in later versions of some browsers. Those later versions are listed as the browser compatibility version for the constructor on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder#Specifications). Does that necessitate any changes to the browser list in `config.toml`? I currently have the earlier versions listed.

We should update the browser list in `config.toml` to include the browsers which have the older constructor behavior, yes 👍 

> Noticed that some packages are listed as third-party packages rather than being checked into the source code. This polyfill code comes directly from [FastestSmallestTextEncoder](https://github.com/anonyco/FastestSmallestTextEncoderDecoder). Is there a preference here between the two options?

It's preferred to use the third-party polyfill system rather than copy and paste the polyfill into this codebase. The reason is to make it simpler to see if any updates to the polyfill have been made, we have automated tooling which will update the third-party polyfill. Would you be able to convert this to using the third-party system we have instead of copy-pasting the polyfill?



> I'm not sure how the folder structure for TextEncoder should work. Should `#encode` be in `polyfills/TextEncoder/encode`? The `TextEncoder` class itself is also being polyfilled here so I figured it was okay to just leave everything in `polyfills/TextEncoder`.

What you've done is correct 👍

JakeChampion:
I ran the tests on CI and one test looks to fail on old Chrome versions. Would you be able to look into this @marshallycheng?
```
Failures:
- chrome/29.0 up to and including chrome/37.0:
   -> TextEncoder TextEncoder #encoding returns 'utf-8' as the encoding
      http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=TextEncoder
      undefined == utf-8
```

marshallycheng:
@JakeChampion absolutely! thanks for pointing that out, i wasn't sure what was failing since the logs in github don't have details. Is there somewhere else I can look to see the more detailed logs?

JakeChampion:
Yeah, the logs on the GitHub Check are a bit hard to deal with which is a shame. I get them by visiting the GitHub Check and clicking the drop-down button which has the three dots and then clicking the "view raw logs" link. This was the link I used for the GitHub Check this timehttps://github.com/Financial-Times/polyfill-library/pull/672/checks?check_run_id=800222532

I will look into improving the contribution workflow by making the GitHub Check comment on the pull request automatically with any test failures and/or linting issues :-)

marshallycheng:
@JakeChampion updated! Removed the `#encoding` test as the library doesnt include it

JakeChampion:
@marshallycheng Let's keep the test and add the missing functionality to the polyfill :-)

marshallycheng:
@JakeChampion sure thing, ill either add the functionality or use a different library when i get the chance

JakeChampion:
@marshallycheng I've made a PR to add TextEncoder.prototype.encoding support --> https://github.com/anonyco/FastestSmallestTextEncoderDecoder/pull/14

marshallycheng:
@JakeChampion Looks great to me! I'll be keeping an eye on that discussion. If there's a new version of the library I will include that one instead

jamesarosen:
It looks like https://github.com/anonyco/FastestSmallestTextEncoderDecoder/pull/14 may have stalled based on philosophical differences. Should we investigate using the much larger but spec-compliant https://github.com/inexorabletash/text-encoding ?

polyfills/TextEncoder/config.toml
```diff
@@ -0,0 +1,30 @@
+aliases = [
+	"TextDecoder",
+	"encoder",
+	"decoder"
```
JakeChampion, line 4:
```suggestion
	"TextDecoder"
```

JakeChampion:
The tests looks to not pass for IE 8, I guess the polyfill does not support IE 8 or we have not declared a dependency on another polyfill which is required.

```
- ie/8.0:
    -> TextEncoder passes the feature detect
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=TextEncoder
       false == true
 - ie/8.0:
    -> TextEncoder TextEncoder has a valid constructor
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=TextEncoder
       'TextEncoder' is undefined
 - ie/8.0:
    -> TextEncoder TextDecoder has a valid constructor
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=TextEncoder
       'TextDecoder' is undefined
 - ie/8.0:
    -> TextEncoder TextDecoder decodes encoded strings
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=TextEncoder
       'TextEncoder' is undefined
```

jamesarosen:
> The tests looks to not pass for IE 8, I guess the polyfill does not support IE 8 or we have not declared a dependency on another polyfill which is required.

https://github.com/anonyco/FastestSmallestTextEncoderDecoder says for IE ≥4 <10

> * Partial polyfill means that Array (or Buffer in NodeJS) will be used instead of Uint8Array/[typedarray].

I don't see anything in our tests that would specifically require `Uint8Array`. Nor any reason IE9 would pass while IE8 would not.

marshallycheng:
I'll take a look when I get the chance next week!

marshallycheng:
@JakeChampion here's an update, not sure what next steps should be. I could try another text encoding library if you're okay with that

I ran `npm run test-polyfills -- --features=TextEncoder`, which successfully ran the tests on browserstack at `localhost:9876` for IE 9 and 10, but failed to run any tests for IE8. 

I also ran `npm run test-polyfills -- --features=document.getElementsByClassName` and tried with a couple other features and had the same results where the karma server hung on `Starting` on IE8. Is this related to `test/polyfills/server.js`?
```
if (isIE8 && !semver.satisfies("8.0.0", config.browsers.ie)) {
        continue;
      }
```

I took a look through the [source](https://github.com/anonyco/FastestSmallestTextEncoderDecoder/blob/master/EncoderDecoderTogether.src.js) for `FastestSmallestTextEncoder` and did not notice anything that would fail on IE8 but not IE9.

JakeChampion:
The code in anonyco/FastestSmallestTextEncoderDecoder does have an IE8 bug, I've fixed it in a local copy and pushed the fix to this pull-request, the tests pass in IE 8 now for me 👍  

The bug is around IE 8's Object.prototype.toString not working when called without a context.
E.G.
```js
var a = Object.prototype.toString;
a();
```
Throws an error in IE 8 when it should return `'[object Undefined]'`

marshallycheng:
Thanks for the help! Great find

PR #668, opened 2020-06-15
FormData : add polyfill

romainmenke:
fixes : https://github.com/Financial-Times/polyfill-library/issues/31

Any feedback is welcome!

- I am pretty sure that the list of dependencies is incomplete.
- What is the policy around tests for external polyfills? The lib seems to have extensive tests, should tests be written here too?

PR #677, opened 2020-06-25
Formdata

isocroft:
This PR enables the **FormData** constructor and instance APIs (e.g. *set()*, *keys()*, *entries()* ) using a polyfill located [here](https://github.com/jimmywarting/FormData)

### DIFF
--- a/polyfills/FormData/config.toml
+++ b/polyfills/FormData/config.toml
@@ -1,0 +1,25 @@
+aliases = [
+  "default-3.3",
+  "default-3.4",
+  "default-3.5",
+  "default-3.6",
+  "default",
+]
+dependencies = [ "Blob", "URLSearchParams", "XMLHttpRequest" ]
+spec = "https://xhr.spec.whatwg.org/#interface-formdata"
+docs = "https://developer.mozilla.org/en-US/docs/Web/API/FormData"
+license = "MIT"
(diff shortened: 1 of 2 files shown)


polyfills/FormData/config.toml
```diff
@@ -0,0 +1,25 @@
...
+chrome = "<50"
+chrome_mob = "<50"
+ie_mob = "10 - *"
+edge = "<18"
+edge_mob = "*"
+firefox = "<45"
+firefox_mob = "<45"
+safari = "<12"
+ie = "<11"
+opera = "<15"
```
JakeChampion, line 25: Thanks for the PR @isocroft.

This currently does not use/install the polyfill. To install the polyfill we need to declare it like this:
```suggestion
opera = "<15"

[install]
module = "formdata-polyfill"
paths = [ "formdata.min.js" ]

```

We will also need to add it as a dependency to the package.json file by running `npm i formdata-polyfill` and updating the .gitignore and .eslintignore files to include `polyfills/FormData/polyfill.js`.

We also want to add some tests in `polyfills/FormData/tests.js` to prove the polyfill works correctly in all the browsers.

JakeChampion:
Closing this in favour of https://github.com/Financial-Times/polyfill-library/pull/668 where the conversation and work seems to be progressing

PR #703, opened 2020-07-06
Fix Map regression introduced in #219

rmja:
This fix resolves an issue where classes are used as Map keys.

The present behavior is that a base class and sub class resolves the same entry. The desired behavior is for them to resolve to different entries.

Test that reproduces the bug and the fix are included.

This regression has made it impossible to use Polyfill.io for aurelia apps in IE.


JakeChampion:
@rmja Thank you for this!

PR #715, opened 2020-07-13
File polyfill

romainmenke:
@JakeChampion this is already a good start for the File polyfill.
I would like to do a self review first tomorrow to make sure I did not miss anything to obvious/basic.
Will poke when this is ready.

### DIFF
--- a/polyfills/File/config.toml
+++ b/polyfills/File/config.toml
@@ -1,0 +1,25 @@
+aliases = []
+dependencies = [
+  "_ESAbstract.CreateMethodProperty",
+  "Object.create",
+  "Object.defineProperty",
+  "Blob"
+]
+notes = [
+  "https://w3c.github.io/FileAPI/#file-section"
+]
+
+[browsers]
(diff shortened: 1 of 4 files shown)


romainmenke:
This is more complicated than expected :)
It is definitely possible to create a polyfill that acts like a `File` but I fear its usefulness would be limited.
The object that is created would be a `Blob` with extra bits attached. This makes it difficult to implement all the other polyfills to actually do something with it.

Currently the `Blob` polyfill also does not implement some methods (`text()`) which again make it harder to have a fully working polyfill.

imho it might be better if a [FormData/File polyfill](https://www.npmjs.com/package/formdata-polyfill) is included by individual projects that need it.
To include it here it should to be fully spec compliant and work nicely together with all the other polyfills here.

If someone with better insights into this wants to take over, I ended up here and will leave that PR open :  https://github.com/romainmenke/polyfill-library/pull/32

@JakeChampion going to close this to reduce PR clutter.

PR #752, opened 2020-07-24
Fix "finally" support for Chrome 63

joshsalverda:
I was able to track down, through a bunch of investigation, that `Promise.prototype.finally` wasn't actually added into Chrome until at least version 63.0.3219

I'm hoping that the main lib supports specifying minor versions this way, otherwise we might have to just change this to `< 64`?

### DIFF
--- a/polyfills/Promise/prototype/finally/config.toml
+++ b/polyfills/Promise/prototype/finally/config.toml
@@ -13,9 +13,9 @@ license = "MIT"
 docs = "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally"
 
 [browsers]
-android = ">4.4 <63"
+android = ">4.4 <64"
 bb = ">10"
-chrome = ">31 <63"
+chrome = ">31 <64"
 firefox = ">28 <58"
 ios_saf = ">7.1 <11.3"
 edge = ">12 <18"


JakeChampion:
Hi @joshsalverda, we unfortunately do not support detecting browser patch versions, we only support major and minor.

joshsalverda:
Thanks for the heads up @JakeChampion, I've updated the pr to use `<64` instead. Also updated the android version since I forgot they switched to chrome eventually as well.

It shouldn't have a huge impact, but let me know if you think otherwise. I don't even know who uses such old chrome versions anymore, but there is at least one person out there, since that's what started me down this rabbit hole.

isocroft:
@joshsalverda this is a very important PR. Do you think you can spare a little more time to see it through ?

joshsalverda:
@isocroft Definitely, if you could give me a slight push in the direction I need to take next? 😅 

Is it just this step that I'm missing? https://github.com/Financial-Times/origami-project-board-action#1-adding-the-secret

The build is complaining about:

>##[error]Parameter token or opts.auth is required

JakeChampion:
> I was able to track down, through a bunch of investigation, that `Promise.prototype.finally` wasn't actually added into Chrome until at least version 63.0.3219

Could you please explain how you came to the conclusion that 63.0.3219 is the version which added the method?
Chrome Status lists it as added in 63.0 -- https://www.chromestatus.com/feature/5704570235781120

My understanding (from speaking with Chrome engineers) of the Chrome version number is that it goes MAJOR.MINOR.BUILD.PATCH and that MAJOR or MINOR is updated for a release to Stable. Which is the reason we only target major and minor. All new features I've seen for Chrome are added in new MAJOR versions, the only modern browser which adds [new features in a MINOR version is Safari](https://gist.github.com/JakeChampion/fab796e59dfd3ff1a831186bf2b21c3d).

joshsalverda:
Hopefully this isn't all craziness, but, basically this started with a Sentry alert saying `.finally is not a function` which I thought was weird, since I use polyfill.io and it should be covered everywhere. The alert mentioned it was Chrome Version: 63.0.3215 which is also weird since the link you mentioned, MDN, caniuse.com, this blog post https://v8.dev/features/promise-finally, everywhere is saying Chrome 63+ supports .finally.

I ended up digging into it a bit since everyone couldn't be wrong, but I wanted to find out why I was seeing this error.

Eventually I ended up in the v8/chromium source code which is where I found this:

The link to the bug for adding finally support is https://bugs.chromium.org/p/v8/issues/detail?id=5967

In there I noticed the commit for "Ship Promise.prototype.finally" was https://chromium.googlesource.com/v8/v8.git/+/9de3fe2ebfe9f6c6993651fc13598c44241e42aa

The next part is all based on that git sha, so if I'm wrong about that being the initial commit for .finally support then this could all be wrong... But the commit does say "Ship" so I'm assuming that is accurate.

I ended up checking out the v8 repo locally so I could trace that sha and ran `git tag --contains 9de3fe2ebfe9f6c6993651fc13598c44241e42aa` and found that the first version (tag) of v8 that actually includes that commit is `6.3.165`

The first chromium tag I could find that references v8 > 6.3.165 was this one: https://chromium.googlesource.com/chromium/src/+/refs/tags/63.0.3219.0

In the DEPS file in there it references v8_revision 03b690ef66a5af748286aeb2be1d229338f8e132 
(See here: https://chromium.googlesource.com/chromium/src/+/refs/tags/63.0.3219.0/DEPS#50)

That 03b690ef66a5af748286aeb2be1d229338f8e132 sha can be traced back to tag 6.3.171 in the v8 repo (using the same tag --contains command above)

The tag right before 3219 is https://chromium.googlesource.com/chromium/src/+/refs/tags/63.0.3218.4

In the DEPS file in there it references v8_revision 6dbf4c4111842cbd19dd920181c78cf48435e288 (https://chromium.googlesource.com/chromium/src/+/refs/tags/63.0.3218.4/DEPS#50)

That 6dbf4c4111842cbd19dd920181c78cf48435e288 is tag 6.3.157 in v8.

So I have to assume that the release for Promise.protoype.finally only showed up in 63.0.3219.0, since that is when the version of the v8 engine that actually shipped finally was added in.

So that's my long explanation... And I hope I'm wrong somehow and just overlooked something simple... Although if that's the case I did waste a whole bunch of time on this for nothing 🤣 

After writing all this out I wonder if it would have just been easier to install a few older versions of chrome on a vm or something but... yeah... 🤷‍♂️

JakeChampion:
@joshsalverda excellent investigative work! I attempted the same search but only got as far as the v8 commit, I couldn't figure out how to link that to a Chromium commit.

Issue #304, opened 2019-09-10
classList polyfill missing .replace() function


lenovouser:
- [developer.mozilla.org/docs/web/api/element/classlist](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) (they have a polyfill that seems to include `.replace()` too there)

 - [developer.mozilla.org/docs/web/api/domtokenlist/replace](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/replace) `DOMTokenList.replace()`

UndefinedBehaviour:
I think this one can be closed now.

PR #775, opened 2020-08-10
polyfill for DOMTokenList.prototype.replace

UndefinedBehaviour:
Hi, this is meant to address issue #304. I tried to follow the conventions you guys use, let me know if I missed something.
The implementation would have been simpler / more performant if I could use DOMTokenList.prototype.value, but that is not in the polyfill yet. If and when DOMTokenList.prototype.value is introduced this can be simplified (but I didn't want to introduce it here as it has its subtleties). The tests were taken from the [codebase of blink](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/external/wpt/dom/nodes/Element-classlist.html).
Any kind of feedback is appreciated!


polyfills/DOMTokenList/prototype/replace/config.toml
```diff
@@ -0,0 +1,14 @@
...
+[browsers]
+android = "<81"
+edge = "<18"
+firefox = "<61"
+ie = "*"
+ios_saf = "<12"
+op_mob = "*"
+opera = "<54"
+safari = "<12"
+samsung_mob = "<9"
```
JakeChampion, line 14: I've added the browsers which were missing and updated ther versions based on the mdn data https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/replace#Browser_compatibility
```suggestion
android = "<67"
bb = "*"
chrome = "<67"
edge = "<18"
edge_mob = "<18"
firefox = "<61"
firefox_mob = "<61"
ie = "*"
ie_mob = "*"
ios_saf = "<12"
op_mini = "*"
op_mob = "<48"
opera = "<54"
safari = "<12"
samsung_mob = "<9"
```

PR #797, opened 2020-08-14
ensure only one pull-request runs the browserstack tests at a single time

JakeChampion:
This should ensure we don't get lots of failing pull-request checks due to browserstack cancelling the test-run due to having too many tests in there queue.

### DIFF
--- a/.github/workflows/test-polyfills.yml
+++ b/.github/workflows/test-polyfills.yml
@@ -5,8 +5,18 @@ on:
       - '**'
       - '!master'
 jobs:
+  check-for-other-pull-requests-running-this-workflow:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Turnstyle
+      uses: softprops/turnstyle@v1
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-polyfills.yml
```diff
@@ -5,6 +5,15 @@ on:
...
+  check-for-other-pull-requests-running-this-workflow:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Turnstyle
+       uses: softprops/turnstyle@v1
+       with:
+        same-branch-only: false
+       env:
+         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   ie:
```
JakeChampion, line 17:
```suggestion
    - name: Turnstyle
      uses: softprops/turnstyle@v1
      with:
        same-branch-only: false
      env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  ie:
```

PR #872, opened 2020-09-23
Update seamless-scroll-polyfill to 1.2.3

magic-akari:


JakeChampion:
Hi @magic-akari, thank you for the pull-request.

We've ran the tests on many browsers and some of them are failing, would you be able to fix these at all?

```
Failures:
 - ie/10.0:
    -> smoothscroll passes the feature detect
       false == true
 - ie/11.0:
    -> smoothscroll passes the feature detect
       false == true
 - ie/9.0:
    -> smoothscroll passes the feature detect
       false == true
 - ie/9.0:
    -> Element.prototype.inert inert set
       Unspecified error.
``` 

To run the tests:
- Build the polyfills with this command --> `npm run build`
- Run the testing server with this command --> `node test/polyfills/server.js`
- Open this url in the browser you want to test in --> http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=smoothscroll

magic-akari:
It seems that the function `Number.isNaN` I used is not listed in the dependencies.
I have fixed it by removing this function.

magic-akari:
I have fixed it. Would you mind rerun the tests?

package.json
```diff
@@ -74,7 +74,7 @@
     "mutationobserver-shim": "^0.3.2",
     "picturefill": "^3.0.1",
     "rimraf": "^3.0.0",
-    "seamless-scroll-polyfill": "^1.0.10",
+    "seamless-scroll-polyfill": "^1.2.3",
```
JakeChampion, line 77:
```suggestion
    "seamless-scroll-polyfill": "1.2.3",
```

PR #884, opened 2020-09-30
Use the new definition steps for Number.IsInteger Number.IsSafeInteger String.fromCodePoint and String.prototype.codePointAt

JakeChampion:


polyfills/Number/isSafeInteger/polyfill.js
```diff
@@ -1,24 +1,13 @@
 /* global CreateMethodProperty, Type, ToInteger */
```
JakeChampion, line 1:
```suggestion
/* global CreateMethodProperty, IsInteger */
```

PR #896, opened 2020-10-13
parallel build script and watch script

romainmenke:
- split up `buildsources.js` into multiple files and documented each bit.
- added `tasks/buildsources/buildsources-child-proc.js` which enables multiple node processes to run simultaneously.
- added `tasks/watchsources` to watch the `polyfills` directory.
- updated `contributing.md` with instructions for `npm run watch`.
- validated output by running the new and original scripts, copied to `__dist` folder and compared with `git diff --no-index polyfills/__dist_parallel/ polyfills/__dist_series`
- kept original build script for now : `npm run build-series`.

Notes :

- `npm run watch` Only rebuilds what is changed and restarts the test server.
- github action VM's seem to have 2 threads available.

see : https://github.com/Financial-Times/polyfill-library/issues/887


tasks/buildsources/build.js
```diff
@@ -0,0 +1,86 @@
...
+		.then(async () => {
+			await makeDirectory(destination);
+
+			const queues = [];
+
+			const maxProc = Math.max(
+				1,
+				Math.min(
+					require("os").cpus().length,
+					process.env["MAXPROCS"] || 9000
```
JakeChampion, line 33: What is `MAXPROCS`? I can't find any information about this environment variable

romainmenke: Forgot to finalise that. Used it during dev to test improvement gain per extra process.

I have now set this at a max of 6. Setting higher does not seen to make it noticeably faster on my machine. 
This has diminishing returns because of the cost of messaging between processes, resource contention, cpu cache misses and lack of cpu boost at high core count usage.

Issue #520, opened 2020-02-29
feature-request: Test that feature detects are not always true or false


JakeChampion:
We've had some feature detects which use IIFE (immediately invoked function expressions) but forgot to invoke them, which meant they were returning a function, and a function is always truthy.

E.G.
```js
'getOwnPropertyNames' in Object && (function() {
    try {
        Object.getOwnPropertyNames(1);
        return true;
    } catch (e) {
        return false;
    }
})
```

Notice that the function is not executed (no `()` at the end).

We need to have a test which checks each `detect.js` file is not always evaluated to true or to false. If we can also have it prove no errors will be thrown, that would be incredible.

I know sonarlint is a tool which can prove an expression will always be true or false, perhaps there is a lighter tool/solution we could use though?

_Originally posted by @JakeChampion in https://github.com/Financial-Times/polyfill-library/issues/218#issuecomment-592972121_

romainmenke:
Have been playing around with this and it produces some interesting results.
Found some broken detects and possibly some polyfill issues : 

https://github.com/romainmenke/polyfill-library/pull/38

Tried a really simplistic approach where I just added `()` after a detect source. This already caught a couple of cases.
Combined with testing for `false` when the polyfill isn't loaded it covers quite a bit of this issue.

Going to poke it a bit more before opening a PR.

PR #898, opened 2020-10-16
more tests and lints for feature detect.js

romainmenke:
resolves : https://github.com/Financial-Times/polyfill-library/issues/520

- add a [linter rule plugin](https://github.com/romainmenke/eslint-plugin-uncalled-iife) to detect uncalled `IIFE` in `detect.js`
- ~~verify that calling `detect.js` as a function throws~~
- fix some `detect.js` files
- fix `DocumentFragment` polyfill

----------

### eslint-plugin-uncalled-iife

I would have preferred to include a custom eslint rule in the `polyfill-library` repo but `eslint` does not support this.
The only way was to create a separate package containing the rule and publish it on npm.
This package is really small and I do not mind maintaining it when eslint introduces breaking changes.

There are cases that can't be caught this way. (named functions that aren't called)

Screenshot 2020-10-16 at 11 15 30

### DocumentFragment

Testing for uncalled `IIFE`'s exposed some issues with the `DocumentFragment ` polyfill.
I do not see how the previous code could have worked, but I might be overlooking something.
If you prefer, I can split this into two PR's.

### test for falsy

I tried adding a test that verifies that the code in `detect.js` returns something falsy in browsers that need the polyfill when the polyfill isn't loaded. In theory this would expose false positives/negatives. In practice this doesn't work :) Too many features would pass a `detect.js` but still need the polyfill. This then required adding a lot of exceptions which makes the test less useful.

### DIFF
--- a/test/polyfills/server.js
+++ b/test/polyfills/server.js
@@ -163,12 +163,6 @@ async function testablePolyfills(isIE8, ua) {
           }).call(window));
         });
 
-        it('throws when called as a function', function() {
-          proclaim["throws"](function () {
-            (${config.detectSource})();
-          });
-        });
-
         ${await readFile(testFile)}
       });`;
       polyfilldata.push({
--- a/test/polyfills/server.js
+++ b/test/polyfills/server.js
@@ -163,6 +163,12 @@ async function testablePolyfills(isIE8, ua) {
           }).call(window));
         });
 
+        it('throws when called as a function', function() {
+          proclaim["throws"](function () {
+            (${config.detectSource})();
+          });
+        });
+
         ${await readFile(testFile)}
       });`;
       polyfilldata.push({
--- a/package-lock.json
+++ b/package-lock.json
@@ -2145,6 +2145,12 @@
         "lodash.zip": "^4.2.0"
       }
     },
+    "eslint-plugin-uncalled-iife": {
+      "version": "1.0.0",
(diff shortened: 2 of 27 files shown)


JakeChampion:
@romainmenke This is superb!

test/polyfills/server.js
```diff
@@ -163,6 +163,12 @@ async function testablePolyfills(isIE8, ua) {
           }).call(window));
         });
 
+        it('throws when called as a function', function() {
+          proclaim["throws"](function () {
+            (${config.detectSource})();
+          });
+        });
+
```
JakeChampion, line 171: Why is this needed? Does it cover things the eslint rule does not?

romainmenke: This catches `function foo(){ return true }` (vs. `function foo(){ return true } && foo()`)

I am not sure if we need or want this test.
These are edge cases and PR review might be a better way to intercept possible issues?

JakeChampion: Yeah, I think we should remove this test and rely on PR review 👍

romainmenke: Cool! done

romainmenke:
@JakeChampion seems a branch was created here : https://github.com/Financial-Times/polyfill-library/tree/more-tests-and-lints-for-feature-detect--passionate-australian-mist-84fb32e692

A bit weird, but can be removed.

JakeChampion:
@romainmenke yep, I need to push the same commits to a branch in the origin (Financial-Times/polyfill-library) in order to have CI to run with the BrowserStack credentials

romainmenke:
Good to know :)

Issue #899, opened 2020-10-17
Symbol.prototype.description polyfill breaks in Safari 13 and earlier


mhassan1:
# Bug Report

## What

The polyfill for `Symbol.prototype.description` fails with error `TypeError: Cannot convert a symbol to a string` in Safari 13 and earlier. This is caused by the type coercion in the function created here: `getInferredName = Function("s", "return { [s]() {} }[s].name;")`.

## Details

Expected Behavior: `getInferredName(Object(Symbol('hi'))) -> '[hi]'`
Actual Behavior: `getInferredName(Object(Symbol('hi'))) -> TypeError (Safari 13)`

To reproduce:
```html
<!-- in Safari 13 and earlier -->
<script src="https://polyfill.io/v3/polyfill.js?features=Symbol.prototype.description"></script>
<script>
console.log(Symbol('hi').description)
</script>
```

PR #900, opened 2020-10-17
rewrite Symbol.prototype.description polyfill so it passes tests

mhassan1:
Resolves https://github.com/Financial-Times/polyfill-library/issues/899

### DIFF
--- a/polyfills/Symbol/prototype/description/polyfill.js
+++ b/polyfills/Symbol/prototype/description/polyfill.js
@@ -21,7 +21,7 @@ if (supportsGetters) {
     var getInferredName;
     try {
       // eslint-disable-next-line no-new-func
-      getInferredName = Function("s", "return { [s]() {} }[s].name;");
+      getInferredName = Function("s", "var v = s.valueOf(); return { [v]() {} }[v].name;");
       // eslint-disable-next-line no-empty
     } catch (e) {}
 
@@ -55,15 +55,55 @@ if (supportsGetters) {
         // 3. Return sym.[[Description]].
         if (supportsInferredNames) {
           var name = getInferredName(sym);
-          if (name === "") {
-            return undefined;
+          if (name !== "") {
+            return name.slice(1, -1); // name.slice('['.length, -']'.length);
           }
-          return name.slice(1, -1); // name.slice('['.length, -']'.length);
+        }
+
+        if (emptySymbolLookup[sym] !== undefined) {
+          return emptySymbolLookup[sym];
         }
 
         var string = sym.toString();
-        return string.slice(7, string.length - 1);
+        if (isSymbolPolyfilled(sym)) {
+          var randomStartIndex = string.lastIndexOf("0.");
+          string = string.slice(10, randomStartIndex);
+        } else {
+          string = string.slice(7, string.length - 1);
+        }
+        if (string === "") {
+          return undefined;
+        }
+        return string;
       }
     });
+
+    function isSymbolPolyfilled(symbol) {
+      return symbol.toString().indexOf("__\x01symbol:") === 0;
+    }
+
+    var emptySymbolLookup = {};
+    var NativeSymbol = self.NativeSymbol = Symbol
+    var SymbolWrapper = function Symbol() {
+      var description = arguments[0];
+      var result = this instanceof SymbolWrapper
+        ? new NativeSymbol(description)
+        // in Edge 13, String(Symbol(undefined)) === 'Symbol(undefined)'
+        : description === undefined ? NativeSymbol() : NativeSymbol(description);
+
+      if (isSymbolPolyfilled(result)) {
+        if (description !== undefined && (description === null || isNaN(description) || String(description) === "")) {
+          emptySymbolLookup[result] = String(description);
+        }
+      } else {
+        if (String(description) === "") {
+          emptySymbolLookup[result] = ""
+        }
+      }
+
+      return result;
+    };
+    SymbolWrapper.prototype = NativeSymbol.prototype;
+    Symbol = SymbolWrapper; // eslint-disable-line no-native-reassign, no-global-assign
   })();
 }
--- a/polyfills/Symbol/prototype/description/polyfill.js
+++ b/polyfills/Symbol/prototype/description/polyfill.js
@@ -1,5 +1,7 @@
 /* global Symbol, Type */
 
+var NativeSymbol = Symbol;
+
 var supportsGetters = (function() {
   try {
     var a = {};
@@ -16,6 +18,7 @@ var supportsGetters = (function() {
     return false;
   }
 })();
+
 if (supportsGetters) {
   (function() {
     var getInferredName;
@@ -83,7 +86,6 @@ if (supportsGetters) {
     }
 
     var emptySymbolLookup = {};
-    var NativeSymbol = self.NativeSymbol = Symbol
     var SymbolWrapper = function Symbol() {
       var description = arguments[0];
       var result = this instanceof SymbolWrapper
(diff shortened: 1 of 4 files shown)


JakeChampion (approved):
Thanks for the pull-request @mhassan1, we really appreicated it :-)
I triggered a run of the tests in our CI environment and here are the test failures which came back.

```
Failures:
 - ie/10.0:
    -> Symbol.prototype.description works with numbers
       undefined === 1
 - ie/10.0:
    -> Symbol.prototype.description works with booleans
       undefined === true
 - ie/10.0:
    -> Symbol.prototype.description works with null
       undefined === null
 - ie/10.0:
    -> Symbol.prototype.description works with arrays
       undefined === 
 - ie/10.0:
    -> Symbol.prototype.description works with objects
       undefined === [object Object]
 - ie/10.0:
    -> Symbol.prototype.description works with regexes
       undefined === /./
 - ie/10.0:
    -> Symbol.prototype.description works with NaNs
       undefined === NaN
 - ie/10.0:
    -> Symbol.prototype.description works with functions
       undefined === function(){}
 - ie/10.0:
    -> Symbol.prototype.description works with an empty string
       undefined === 
 - ie/11.0:
    -> Symbol.prototype.description works with numbers
       undefined === 1
 - ie/11.0:
    -> Symbol.prototype.description works with booleans
       undefined === true
 - ie/11.0:
    -> Symbol.prototype.description works with null
       undefined === null
 - ie/11.0:
    -> Symbol.prototype.description works with arrays
       undefined === 
 - ie/11.0:
    -> Symbol.prototype.description works with objects
       undefined === [object Object]
 - ie/11.0:
    -> Symbol.prototype.description works with regexes
       undefined === /./
 - ie/11.0:
    -> Symbol.prototype.description works with NaNs
       undefined === NaN
 - ie/11.0:
    -> Symbol.prototype.description works with functions
       undefined === function(){}
 - ie/11.0:
    -> Symbol.prototype.description works with an empty string
       undefined === 
 - ie/9.0:
    -> Symbol.prototype.description works with numbers
       undefined === 1
 - ie/9.0:
    -> Symbol.prototype.description works with booleans
       undefined === true
 - ie/9.0:
    -> Symbol.prototype.description works with null
       undefined === null
 - ie/9.0:
    -> Symbol.prototype.description works with arrays
       undefined === 
 - ie/9.0:
    -> Symbol.prototype.description works with objects
       undefined === [object Object]
 - ie/9.0:
    -> Symbol.prototype.description works with regexes
       undefined === /./
 - ie/9.0:
    -> Symbol.prototype.description works with NaNs
       undefined === NaN
 - ie/9.0:
    -> Symbol.prototype.description works with functions
       undefined === function(){}
 - ie/9.0:
    -> Symbol.prototype.description works with an empty string
       undefined === 
```

You can run these tests yourself by running `npm run watch` and then opening http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Symbol.prototype.description in the browser you want to tests against.

mhassan1:
I believe I've addressed that issue, although it turned out to be a bit more of a rabbit hole than I expected. Let me know if this doesn't look right.

JakeChampion:
@mhassan1 hmm, it look not right for some reason.
Here are the tests which now fail:
```
Failures:
 - ie/10.0:
    -> Map Map implements .set()
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Map
       Unable to get property 'name' of undefined or null reference
 - ie/11.0:
    -> Map Map implements .set()
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Map
       Unable to get property 'name' of undefined or null reference
 - ie/8.0:
    -> Map Map implements .set()
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Map
       'constructor.name' is null or not an object
 - ie/9.0:
    -> Map Map implements .set()
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Map
       Unable to get value of the property 'name': object is null or undefined
```

mhassan1:
Thanks for the heads up. Those tests should now be passing.

polyfills/_ESAbstract/Type/polyfill.js
```diff
@@ -15,7 +15,7 @@ function Type(x) { // eslint-disable-line no-unused-vars
 			// typeof null is 'object'
 			if (x === null) return 'null';
 			// Polyfill.io - This is here because a Symbol polyfill will have a typeof `object`.
-			if ('Symbol' in self && (x instanceof self.Symbol || x.constructor === self.Symbol)) return 'symbol';
+			if ('Symbol' in self && (x instanceof self.Symbol || x.constructor === self.Symbol || (x.constructor && x.constructor.name === 'Symbol'))) return 'symbol';
```
JakeChampion, line 18: using `x.constructor.name` means that we need to include `Function.prototype.name` as a dependency for this file.

We will need to add `Function.prototype.name` to the `polyfills/_ESAbstract/Type/config.toml` file's `dependencies` section

JakeChampion:
Thank you for updating it @mhassan1 
I ran it in CI once more and it still has some failing tests, would you be able to look into those at all?

The CI logs are [here](https://github.com/Financial-Times/polyfill-library/runs/1318261107?check_suite_focus=true)

I think the quickest way to find the issue would be to run the Symbol tests in the oldest vesion of Chrome they are currently failing in, which is Chrome 35.

If you run `npm run watch`, you can then run the [Symbols tests](http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Symbol) in Chrome 35 👍

polyfills/_ESAbstract/Type/polyfill.js
mhassan1: I've reverted that change. Instead, we can check for `x instanceof self.NativeSymbol`

JakeChampion:
Nice @mhassan1!

We've now only got two tests for Symbol.prototype.description which are failing:

```
 - chrome/49.0 - chrome/69.0:
    -> Symbol.prototype.description works with undefined
        === undefined
    -> Symbol.prototype.description works with no input
        === undefined
    http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Symbol.prototype.description
```

JakeChampion:
@mhassan1 only one failing test in two browser versions now 🙌 
```
Failures:
 - chrome/49.0:
    -> Symbol.prototype.description works with arrays
       undefined === 
 - chrome/50.0:
    -> Symbol.prototype.description works with arrays
       undefined === 
```

polyfills/Symbol/prototype/description/polyfill.js
```diff
@@ -55,15 +55,55 @@ if (supportsGetters) {
...
+        return string;
       }
     });
+
+    function isSymbolPolyfilled(symbol) {
+      return symbol.toString().indexOf("__\x01symbol:") === 0;
+    }
+
+    var emptySymbolLookup = {};
+    var NativeSymbol = self.NativeSymbol = Symbol
```
JakeChampion, line 86: This is adding a new global variable named `NativeSymbol`, I'm not sure we should do that as it is not part of a specification we are polyfilling

mhassan1: Understood. I could not see any other way for `polyfills/_ESAbstract/Type/polyfill.js` to identify whether an object is an instance of a Symbol, since the `Symbol.prototype.description` wraps `NativeSymbol` in a new global `Symbol` (which is a pattern I took from [core-js](https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.symbol.description.js).

Originally, I was thinking we could add a check for `x.constructor.name` to the `Type` function, but that seemed strange, since someone could create a constructor called `Symbol` that doesn't actually refer to a Symbol.

Other ideas:
* Instead of attaching `NativeSymbol` to global, attach it to `Symbol` and amend `Type` to check `x instanceof Symbol` or `x instanceof Symbol.NativeSymbol`, although this also modifies `Symbol` in an unspecified way

I'm certainly open to other ideas.

JakeChampion: I made NativeSymbol a variable which is in the same scope as the `Type` function when the polyfill bundle is built, this way it doesn't need to be a global variable and everything still works :-)

Thanks for this excellent work by the way! 🙇

JakeChampion (approved):
Top notch work here @mhassan1, thank you so much for this

Issue #902, opened 2020-10-24
[Bug] While Smoothscroll detecting user scrolls to the top of the page.


kholiavko-roman:
## What

While Smoothscroll detecting user scrolls to the top of the page.

## Details
As you can see ton the demo 
![ezgif-6-5b0cc94026f1](https://user-images.githubusercontent.com/16399378/97088368-f20f4a00-1638-11eb-8a43-44933e0853f1.gif)
while page is loading user can scroll but after script was loaded it will scroll page to the top.
https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/smoothscroll/detect.js#L20

Browser: Safari
Device: iPhone 7, iOS 11

PR #903, opened 2020-10-24
FIX: smoothscroll detect.

kholiavko-roman:
FIX: smoothscroll detect.

Resolves https://github.com/Financial-Times/polyfill-library/issues/902

I think calling `document.body.scrollTo(scrollOptions)` is not necessary here 
https://github.com/Financial-Times/polyfill-library/compare/master...kholiavko-roman:fix-smoothscroll-detect?expand=1#diff-87ce488040c638608dfb1662f22ac99e6fa071348163b26fb3682916711eeebeL20
because it actually scroll page to the top that makes UX worse.
Just `'scrollBehavior' in document.documentElement.style` works fine.


JakeChampion:
Thanks for the PR @kholiavko-roman, in order for us to accept and merge this we need you to sign our [Contributor License Agreement](https://cla-assistant.io/Financial-Times/polyfill-library?pullRequest=903). Once that is signed and the tests pass in CI we can merge and release this :-)

kholiavko-roman:
@JakeChampion I accepted the licence but looks like it test for IE was failed. I will take a look tomorrow.

kholiavko-roman:
@JakeChampion could you please take a look at test on ie browsers ? https://github.com/Financial-Times/polyfill-library/pull/903/checks?check_run_id=1309541666
As far as I understand tests expected that in IE 9-11 smoothscroll feature detect must return true (it means that it works) but I tested in real browser IE 11 and looks like this browser doesn't support this feature, here is a video:
ezgif-1-f8a7c822394d

romainmenke:
@kholiavko-roman As far as I understand, the polyfill does not make this true : 
`('document' in self && 'documentElement' in self.document && 'style' in self.document.documentElement && 'scrollBehavior' in document.documentElement.style)`

That is where the original detect code comes in.
I added a modified version that scrolls a newly created div element.
This fixes the attached issue without changing the logic of `detect.js` too much.

Would this work?

```js
('document' in self && 'documentElement' in self.document && 'style' in self.document.documentElement && 'scrollBehavior' in document.documentElement.style) || (function () {
	try {
		var supportsSmoothScroll = false;
		
		var scrollOptions = {
			top: 1,
			left: 0
		};
		
		Object.defineProperty(scrollOptions, 'behavior', {
			get: function () {
				supportsSmoothScroll = true;
				return 'smooth';
			},
			
			// Ensure this property lasts through cloning / destructuring:
			enumerable: true
		});
		
		var a = document.createElement('div');
		var b = document.createElement('div');
		a.setAttribute('style', 'height: 1px');
		b.setAttribute('style', 'height: 2px');
		
		a.appendChild(b);
		a.scrollTo(scrollOptions);
		
		return supportsSmoothScroll;
	} catch (e) {
		return false;
	}
})();
```

JakeChampion:
@romainmenke does that scroll the page to the top after the detect has run? That's the bug that @kholiavko-roman is trying to fix here I believe

romainmenke:
@JakeChampion If I run it on any page (scrolled down) it doesn't have any effect on the scroll position. (only the second part of the detector)
Did update the `style` attribute in the snippet to include `overflow: scroll;` Doesn't seem to have much effect on the outcome, but is technically more correct.

kholiavko-roman:
@romainmenke yes this code in IE returns `false`
`('document' in self && 'documentElement' in self.document && 'style' in self.document.documentElement && 'scrollBehavior' in document.documentElement.style)`
If it `false` It means that that browser doesn't support this feature, correct ?


By the way your code also return `false` in Internet explorer 🤔
image

JakeChampion:
> @romainmenke yes this code in IE returns `false`
> `('document' in self && 'documentElement' in self.document && 'style' in self.document.documentElement && 'scrollBehavior' in document.documentElement.style)`
> If it `false` It means that that browser doesn't support this feature, correct ?
> 
> By the way your code also return `false` in Internet explorer 🤔
> image

That's correct.

The `detect.js` file should evaluate to `true` after the polyfill has been applied.

kholiavko-roman:
@JakeChampion @romainmenke 
Interesting thing, I just tested original solution from master https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/smoothscroll/detect.js
And it also return false in IE11.
image

Maybe something wrong with tests ?

romainmenke:
@kholiavko-roman returning `false` is actually the goal for all browsers that do not support the feature, or before loading the polyfill.

The only issue with the original `detect.js` was that it changed the page scroll position. It did always correctly return `true` or `false`.

kholiavko-roman:
@romainmenke Maybe I confused a little bit from tests output
`false == true` which one is actual result and which  one is expected result ? 🧐
image

I thought that the second (true) value was expected result because code after my changes still returns `false` in IE 
image

romainmenke:
@kholiavko-roman Sorry, I just now realised what is confusing 🙂 
The tests in **CI** run with the polyfill loaded when needed. So in CI `detect.js` must always return `true`.

When running it manually in a browser that needs the polyfill it should be `false` without polyfill.

So the test results are reporting that `false` was returned when `true` was expected (polyfill is loaded here).
When you run it manually in IE11 it returns `false` as there is no polyfill here.

You can test this by running `npm run watch` and then visiting :

with polyfill : http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=smoothscroll
no polyfill : http://bs-local.com:9876/test?includePolyfills=no&always=no&feature=smoothscroll

polyfills/smoothscroll/detect.js
```diff
@@ -1,26 +1 @@
...
-
-        document.body.scrollTo(scrollOptions);
-
-        return supportsSmoothScroll;
-    } catch (e) {
-        return false;
-    }
-})();
\ No newline at end of file
+'document' in self && 'documentElement' in self.document && 'style' in self.document.documentElement && 'scrollBehavior' in document.documentElement.style;
```
JakeChampion, line 1: Suggestion from @romainmenke -- https://github.com/Financial-Times/polyfill-library/pull/903#issuecomment-716687789

```suggestion
('document' in self && 'documentElement' in self.document && 'style' in self.document.documentElement && 'scrollBehavior' in document.documentElement.style) || (function () {
	try {
		var supportsSmoothScroll = false;
		
		var scrollOptions = {
			top: 1,
			left: 0
		};
		
		Object.defineProperty(scrollOptions, 'behavior', {
			get: function () {
				supportsSmoothScroll = true;
				return 'smooth';
			},
			
			// Ensure this property lasts through cloning / destructuring:
			enumerable: true
		});
		
		var a = document.createElement('DIV');
		var b = document.createElement('DIV');
		a.setAttribute('style', 'height: 1px; overflow: scroll;');
		b.setAttribute('style', 'height: 2px; overflow: scroll;');
		
		a.appendChild(b);
		a.scrollTo(scrollOptions);
		
		return supportsSmoothScroll;
	} catch (e) {
		return false;
	}
})();
```

kholiavko-roman:
@JakeChampion @romainmenke Thanks for explanation! And looks like your suggestion works fine, it passed CI tests and I tested on real device it doesn't scroll.

PR #911, opened 2020-10-31
fix : test director skips single feature

romainmenke:
Handlebars was used before to turn the feature list into a comma separated string which resulted in a trailing comma.
`Array.join` is now used which does not have a trailing comma.

--------

ci fails in chrome on `AbortController`. This was never tested before as it is the first feature. (order is inversed twice I think)

--------

ci running now and seems to work better. Replace fetch calls to `https://httpstat.us/200?sleep=1000` with an internal "sleep" endpoint.

We had an issue some time ago where older OS's on Browserstack couldn't connect over `https` anymore because of the Sectigo cert expiry. My best guess is that this was the same issue.

Will run CI again to make sure it isn't just flacky and accidentally passing now.

--------

all green on this end

### DIFF
--- a/polyfills/AbortController/tests.js
+++ b/polyfills/AbortController/tests.js
@@ -1,5 +1,12 @@
 /* globals proclaim */
 
+beforeEach(function (done) {
+    // 
+    setTimeout(function () {
+        done();
+    }, 50);
+})
+
 describe('AbortSignal', function () {
     // Because these tests make use of the network, they can be unreliable and fail for reasons outside of the polyfills concern such as a Network Timeout.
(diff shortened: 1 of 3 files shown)


1.
```diff
@@ -0,0 +1,148 @@
+*   Trying 104.31.78.83...
```
JakeChampion, line 1: @romainmenke do we need this?

romainmenke: ugh no,  was looking at supported tls versions `httpstat`.
missed this during cleanup

romainmenke: p.s. force pushed as that file should not be in the git history.

PR #912, opened 2020-10-31
attempt at less flaky ResizeObserver tests

romainmenke:
Noticed that `ResizeObserver` tests sometimes fail. Did not find one root cause but is most likely caused by `requestAnimationFrame` not firing reliably (load on Browserstack vm?).

Extending timeouts and adding a few retries should reduce the number of failures.

--------------

Also added a per commit cache of the `polyfills/__dist` folder.
This is used only to share the build result between steps in a single run.

Full test suite now takes ±20 minutes (down from ±25 minutes)


JakeChampion (approved):
Excellent work on speeding up CI runs even more @romainmenke  🙌

romainmenke:
Happy to help!
I do like my feedback cycles short :)

PR #917, opened 2020-11-28
bump eslint -> 7.14.0

romainmenke:
- bump eslint to latest major
- bump eslint depedencies
- fixed linting errors
- disabled two rules :
  - `unicorn/prefer-optional-catch-binding` (needs ES10/2019 environment or additional build steps)
  - `unicorn/prefer-ternary` (code would often be less readable)
- locally ignored `no-reduce` rule :
  - imho code with reduce is harder to understand, so best avoided going forward?
  - existing code can stay as is to keep this change as small as possible

[Lint results](https://github.com/romainmenke/polyfill-library/runs/1467203607?check_suite_focus=true)
[Test results](https://github.com/romainmenke/polyfill-library/runs/1467249648?check_suite_focus=true)


.eslintrc.js
```diff
@@ -12,5 +12,8 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2018
   },
-  rules: {}
+  rules: {
+    "unicorn/prefer-optional-catch-binding": "off",
+    "unicorn/prefer-ternary": "off",
```
JakeChampion, line 17: Definitely agree here, ternaries are usually less readable to me also 👍

JakeChampion (approved):

PR #918, opened 2020-11-29
fail tests on missing detect source with better feedback

romainmenke:
Result when deleting `detect.js` for a polyfill : 

Screenshot 2020-11-29 at 19 55 08

This should make it easier to spot this kind of issue : https://github.com/Financial-Times/polyfill-library/pull/913#issuecomment-734773203

There the test has a syntax error in older browsers when `detect.js` is missing (`return ();`).

[CI Results](https://github.com/romainmenke/polyfill-library/runs/1470577858)


JakeChampion (approved):
Great investigative work here @romainmenke 🙌

PR #921, opened 2020-12-03
Serve DOMRect polyfill to Edge versions < 79

callum:
Info taken from here: https://github.com/mdn/browser-compat-data/blob/2084025a3264621792d82cc5f76f397bf9727335/api/DOMRect.json


JakeChampion (approved):

callum:
Thanks for the approval @JakeChampion - when is realistic to expect this to be available through https://polyfill.io/v3/polyfill.min.js?

callum:
Hi @JakeChampion, should this be up on polyfill.io now? I've just checked and it's not coming through for Edge 18.

JakeChampion:
@callum No, the release cycle for polyfill.io is not as fast as the release cycle for polyfill-library.

here is a pr which updates polyfill.io -- https://github.com/Financial-Times/polyfill-service/pull/2461

PR #931, opened 2020-12-06
various updates to browser configs

romainmenke:
_more to come_


romainmenke:
update : did not encounter any other config that needed updating. (or was combined with a polyfill fix.)

PR #935, opened 2020-12-07
fix Symbol.prototype.description

romainmenke:
_this is the big one_

I am not very confident about this code and this change might introduce new bugs :/
What I wanted to do :

- only ship the entire `Symbol` polyfill to browsers that really need it. Not to every browser that needs `Symbol.prototype.description`
- still have a working polyfill for `Symbol.prototype.description` in all supported browsers

The only way I could think of to do that, was by using a `Proxy` in modern browsers.
This does introduces duplicate logic and duplicate tests.

I am definitely not an expert on `Symbol` and might have overlooked nuances when mangling the code for both polyfills.

_ci is still running for this one : [reference branch](https://github.com/romainmenke/polyfill-library/runs/1513832631?check_suite_focus=true)_


romainmenke:
@JakeChampion I forgot to finalise the browsers list in `config.toml` here.
When double checking everything I noticed that Safari 9 doesn't have `Proxy`.

So I tweaked the polyfill to have a minimal implementation for browsers without `Proxy`.

PR #936, opened 2020-12-08
fix String.prototype.replaceAll

romainmenke:

PR #937, opened 2020-12-10
fix String.prototype.replaceAll

romainmenke:
_supersedes :_ https://github.com/Financial-Times/polyfill-library/pull/936

I had to update a lot of tests.

I duplicated the tests which erased `RegExp.prototype[@@replace]` to make sure we also test the behaviour when it does exist. This will help later when we add a polyfill for `RegExp.prototype[@@replace]`

-----------

`context("functional replacer", function() {`

That line caused the tests to be skipped without output.

After changing that I took the actual output and set it as the expected 😏 
I then verified this in a modern browser which doesn't need the polyfill.
_so not entirely the best way of working._

----------

We might be able to get support for IE8 with this snippet :

```js
		// 2.b. If isRegExp is true, then
		if (isRegExp) {
			// 2.b.i. Let flags be ? Get(searchValue, "flags").
			var flags = Get(searchValue, "flags");
			
			// 2.b.iii. If ? ToString(flags) does not contain "g", throw a TypeError exception.
			if (flags === null && searchValue.global !== true) {
				throw TypeError('');
			} else {
				// 2.b.ii. Perform ? RequireObjectCoercible(flags).
				RequireObjectCoercible(flags)
				// 2.b.iii. If ? ToString(flags) does not contain "g", throw a TypeError exception.
				if (ToString(flags).indexOf('g') === -1) {
					throw TypeError('');
				}
			}
		}
```

IE8 doesn't get the `flags` polyfill but does have `global` as a property.

-----

[reference branch run](https://github.com/romainmenke/polyfill-library/runs/1533807106?check_suite_focus=true)

had a flaky test but running again now, will check results tomorrow.


JakeChampion:
@romainmenke nice idea for IE8, I like it 👍

polyfills/String/prototype/replaceAll/config.toml
```diff
@@ -69,10 +71,14 @@ edge_mob = "<85"
...
+ie = ">=9"
+ie_mob = ">=9"
 opera = "<71"
 op_mini = "*"
 safari = "<13.1"
 firefox_mob = "<79"
 samsung_mob = "*"
+
+notes = [
+  "Polyfill for RegExp.prototype[@@replace] is missing. This polyfill won't work fully as expected untill that polyfill has been implemented."
```
JakeChampion, line 83: Good note 👍

romainmenke:
Updated and reference branch passes CI for IE8 :

https://github.com/romainmenke/polyfill-library/actions/runs/414867266

There was one quirk in IE8 :

```js
// IE8 reorders flags when converting a RegExp to string.
if ((/./gi).toString() === "/./ig") {
```

PR #942, opened 2020-12-14
add "es2015" to the aliases of String.fromCodePoint

hkano:
I think there should be `"es2015"` in the `aliases` of `String.fromCodePoint`, what do you think?
Please confirm.


JakeChampion (approved):

JakeChampion:
Thanks for this @hkano

PR #944, opened 2020-12-16
Update smoothscroll detect logic

bricksroo:
I noticed the `smoothscroll` polyfill wasn't patching properly in Safari when using the polyfill-service. The first part of the detect logic returns `false` (indicating the need for the polyfill), while the second returns `true`. However, when wrapping the detect logic with `!()` for the `?flags=gated` case, it evaluates to `false` due to the `||` operator. 

Screen Shot 2020-12-15 at 11 16 06 PM

This PR updates the logic from `||` to `&&`.

Please let me know if there is any more info I can provide or things I forgot to address with this PR.

### DIFF
--- a/polyfills/smoothscroll/detect.js
+++ b/polyfills/smoothscroll/detect.js
@@ -1,4 +1,18 @@
-('document' in self && 'documentElement' in self.document && 'style' in self.document.documentElement && 'scrollBehavior' in document.documentElement.style) && (function () {
+(function () {
+  var supportsScrollBehaviorViaCss = ('document' in self && 'documentElement' in self.document && 'style' in self.document.documentElement && 'scrollBehavior' in document.documentElement.style);
+
+  if (supportsScrollBehaviorViaCss) {
+    return true;
+  }
+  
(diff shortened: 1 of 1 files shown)


JakeChampion (approved):

JakeChampion:
@bricksroo If the second part of the detect is returning true that means that Safari supports scroll-behaviors. Which version of Safari are you referring to by the way?

We have a long thread in a previous pull-request about why this uses an or operator (`||`), it is to avoid unnecessarily making the page scroll for browsers which do support scroll-behaviors -- https://github.com/Financial-Times/polyfill-library/pull/903

bricksroo:
@JakeChampion This is in the latest version of safari on macOS. 
Screen Shot 2020-12-16 at 9 08 38 AM

After a bit more testing, it seems that safari doesn't catch when trying the `scrollTo(scrollOptions)` with `behavior: 'smooth'`. It successfully scrolls the element, however without the actual "smooth" behavior. Safari seems to essentially ignore the `behavior` value of `scrollOptions`. [Here is MDN's live example for ScrollToOptions](https://mdn.github.io/dom-examples/scrolltooptions/).

bricksroo:
Thank you for taking a look. Appreciate you!

Issue #667, opened 2020-06-11
TypedArray polyfills are included for every browser target


connorjclark:
TypeArray marks all browsers as `*`: https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/_TypedArray/config.toml#L19

These polyfills were added in #1374.

In https://github.com/Financial-Times/polyfill-library/commit/efab9f1ea7a57832595c4025b35328179776d33e#diff-0ad6891d5cf8b4686627e337d7ce5f4c it was all changed to `*`. Note that this is the only file in this commit that version information was lost. Might it have been by mistake?

PR #945, opened 2020-12-16
Update targeting for TypedArray polyfills

JakeChampion:
The numbers were taken from MDN -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#Browser_compatibility

Fixes https://github.com/Financial-Times/polyfill-library/issues/667

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,7 +5,7 @@ polyfills/__dist/**
 # We do not ignore the config.json, detect.js or tests.js
 # because we do control their implementation.
 
-polyfills/_TypedArray/polyfill.js
+polyfills/ArrayBuffer/polyfill.js
 polyfills/atob/polyfill.js
 polyfills/AudioContext/polyfill.js
 polyfills/EventSource/polyfill.js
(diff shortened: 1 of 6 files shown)


romainmenke:
@JakeChampion taking a look at this, no idea why the error is empty.

JakeChampion:
> @JakeChampion taking a look at this, no idea why the error is empty.

Thanks, I got side-tracked by other stuff at work and haven't looked into why this fails yet

romainmenke:
> Each internal polyfill needs to target all supported browsers at all versions.

There is a check for Polyfills that start with `_`.
If so all browsers must have `*` in the config.

_Will patch the build script to have better errors._

JakeChampion:
> > Each internal polyfill needs to target all supported browsers at all versions.
> 
> There is a check for Polyfills that start with `_`.
> If so all browsers must have `*` in the config.
> 
> _Will patch the build script to have better errors._

I thought that might be the issue, I'll need to rename the polyfill to not be internal anymore as we want to get correct browser targeting information

PR #924, opened 2020-12-04
When testing, include all polyfills which target the browser rather than only the polyfill being tested

JakeChampion:

PR #951, opened 2020-12-20
compress test server responses and add test matrix with partial retries

romainmenke:
Since merging #924 test runs fail a lot more often.

I ran tests a bunch of times to see if there was any pattern and it seems that if the first request from `test-runner.handlebars` is a large bundle there is a higher chance of a failure.

Failure rate on iOS Safari 10.x was about 60%.
This alone was enough to make CI frustratingly flaky.

- `Intl.*` test runs fail often.
- Any run with all polyfills loaded fails more often.
- Not every browser is affected equally, older iOS is particularly sensitive to this.

In the end I found that compressing responses from `server.js` helps enough.
This change does add a new dependency : https://www.npmjs.com/package/compression

I also converted the github action to a matrix as it has the ability to continue with other jobs when one fails while still having an overal failure of the workflow. A test run "witness" is "recorded" to skip jobs when a retry is needed. This is done by caching a file (which only happens when the job succeeds) and checking for existence of this cache. 

This also makes it easier to get a good overal view of which browsers have issues after a change instead of first fixing IE, retrying CI, then fixing Chrome,...

_the name "witness" comes from make, but I can't find the reference anymore._

Android fails, Chrome still runs :

Screenshot 2020-12-20 at 21 03 54

IE is skipped on retry :

Screenshot 2020-12-20 at 21 21 02


.github/workflows/test-polyfills.yml
```diff
@@ -21,172 +21,48 @@ jobs:
...
-    - run: npm run build
-      if: steps.cache-dist.outputs.cache-hit != 'true'
-    - name: Test on safari
-      run: node ./test/polyfills/server.js & node ./test/polyfills/remotetest.js targeted director browser=safari
+        path: $GITHUB_WORKSPACE/witness-${{ matrix.browser }}
+        key: witness--${{ matrix.browser }}--${{ env.commit-sha }}
+
+    - name: Test ${{ matrix.browser }}
+      run: node ./test/polyfills/server.js & node ./test/polyfills/remotetest.js targeted director browser=${{ matrix.browser }}
+      if: steps.witness.outputs.cache-hit != 'true'
```
JakeChampion, line 65: This is very smart, nice idea @romainmenke 🙌

JakeChampion (approved):

PR #983, opened 2021-01-22
URLSearchParams: firefox needs polyfill up to 53

romainmenke:
`URLSearchParams.prototype.toString` should have been implemented in Firefox 29 according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

In practice you get `[Object Object]` URL encoded.
From Firefox 54 this works natively.

`URLSearchParams.prototype.sort` was also implemented in Firefox 54 which just happens to align with this fix.

-----------

**Browser log in Firefox 53 :**

Screenshot 2021-01-22 at 19 41 00

**Test run without polyfills in Firefox 53 :**

Screenshot 2021-01-22 at 19 42 35

**Test results for which show the range where `URLSearchParams.prototype.toString` works :**

Screenshot 2021-01-22 at 19 45 55


JakeChampion:
Thanks @romainmenke, running on CI now 👍

JakeChampion (approved):

PR #987, opened 2021-02-01
DocumentFragment : update browsers config

romainmenke:
MDN compat data for Edge is incorrect.
Only in Edge 17 does `new DocumentFragment()` work.

IE has no support at all.

------------

Test for `new DocumentFragment()` :

_edge 15 passed for core-web as it is injected there but gated.
the edge 16 bundle does not include the polyfill._

Screenshot 2021-02-01 at 21 35 40


romainmenke:
@JakeChampion found some more configs that need updating here : [see](https://github.com/mdn/browser-compat-data/pull/8925)

Will revisit this tomorrow.

romainmenke:
Test results : (search for DocumentFragment) https://romainmenke.github.io/web-tests/

- Chrome tests here in polyfill-library only start from Chrome 29 but the polyfill is confirmed to work up to Chrome 15 (oldest I can test)
- Safari 6.2 added support for `DocumentFragment constructor`. MDN browser version info doesn't have an entry for Safari 6.2 so I added it there as `from 7`. (I am unsure what the general best practices are, maybe they should both be `6.2`)
- all other browser versions where taken from engine equivalents as found here : https://gist.github.com/romainmenke/2a2a7cd61e5242d929bdcc1cece94f87 
(e.g. Android 4.4 uses Blink 30 which seems to be the first Android Webview update after Blink 29)

JakeChampion (approved):
Great work as usual @romainmenke, thank you for building your testing matrix, it's proving very useful!

PR #990, opened 2021-02-11
Polyfill for : URL.prototype.toJSON

romainmenke:
Notes :

MDN and Spec differ slightly in wording but I am unsure if this also implies a difference in implementation :

- [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URL/toJSON) : `toJSON` has the same result as `toString`
- [Spec](https://url.spec.whatwg.org/#dom-url-tojson) : `toJSON` has the same result as `href`


JakeChampion (approved):
Nice work :-)

PR #991, opened 2021-02-14
Polyfill for : HTMLSelectElement.prototype.selectedOptions

romainmenke:
fixes : #257

-----

_CI all green on my end_

PR #997, opened 2021-02-20
partial tests in ci

romainmenke:
## Changes for tests in CI :

1. testing polyfill combinations is no longer done per run, moved to a scheduled job on `master`.
2. only needed polyfills and browsers are tested if a diff against `master` contains few changes only in `polyfills/...`.

### 1 : Scheduled job for polyfill combinations.

These tests are still valuable as they revealed a lot of issues initially.
New errors however are rare and these tests should just guard against issues creeping in over time.

Testing once a week on master should be sufficient to keep the codebase healthy.

The caching mechanism in place to speed up retries will also help here.
If there were no changes in the past week this job will skip each step.

### 2 : Partial tests.

1. A diff is taken when `test-modified-only` is used as an argument for `node ./test/polyfills/remotetest.js`.
2. This compares the current commit and branch against `master` on `https://github.com/Financial-Times/polyfill-library.git`
It doesn't take `origin/master` to make it easier for contributors who fork and open pr's. Their CI should not test against their own `master` but against `upstream/master`
3. This file diff is then analysed to first determine if only polyfills were updated. If there are other file changes the full test suite is run. If there are too many polyfill changes the full test suite is also run.
4. The full dependants graph for each modified polyfill is taken to construct the list of features to test.
5. The list of browsers from Browserstack is filtered to only contain browsers which need at least one polyfill that will be tested.

------

## Notes :

This is a rather big change and I would like to revisit it in the next couple of days with fresh eyes.
Any feedback in the mean time is more than welcome!

If there is a reason not to implement a feature like this at all, that is totally fine too! 🙂 

------

This PR would be a good validation of this change as it's only needed in a few old browser : https://github.com/Financial-Times/polyfill-library/pull/991


test/polyfills/test-job.js
```diff
@@ -172,6 +172,6 @@ module.exports = class TestJob {
   }
 
   get configForLog() {
-    return `${this.mode.padEnd(" ", 8)} / ${(this.polyfillCombinations ? 'combined' : '        ')} / ${(this.shard ? 'shard ' + this.shard + '  ' : 'shard n/a')}`;
+    return `${this.mode.padEnd(" ", 8)} / ${(this.polyfillCombinations ? 'combined' : '        ')}${(this.shard ? ' / shard ' + this.shard + '  ' : '')}`;
```
romainmenke, line 175: Only IE8/9 need this, for all other tests it is just bloating the logs.

romainmenke:
@JakeChampion Also added some unit tests to validate the logic and at this point I am generally happy with the state of this change.

JakeChampion (approved):
this is great!

PR #1005, opened 2021-02-28
Element.prototype.getAttributeNames

romainmenke:
Spec notes that the list is ordered but this is different for Gecko/Trident/Webkit.
So tests can't do a deep equal array check.

Spec also indicates that the list can contain duplicates but I found no reference to this behaviour outside of spec/mdn and could not replicate this. (using `innerHTML` to set the same attribute name twice doesn't result in duplicate values)

-----------------

_wanted to validate and test CI and this ran smoothly in `13m 8s`_


polyfills/Element/prototype/getAttributeNames/config.toml
```diff
@@ -0,0 +1,32 @@
...
+#### Browser compatibility
+[browsers]
+android = "<61"
+bb = "*"
+chrome = "<61"
+edge = "<18"
+edge_mob = "<18"
+firefox = "<45"
+ios_chr = "<10.1"
+ios_saf = "<10.1"
```
JakeChampion, line 25: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttributeNames#browser_compatibility) states iOS safari added support in 10.3

@romainmenke did you test in iOS 10.2 and/or 10.3?

JakeChampion:
> Spec notes that the list is ordered but this is different for Gecko/Trident/Webkit.

I think the web platform tests has a set of tests for this behavior and has recorded a bug against Gecko for their current behaviour -- https://wpt.fyi/results/dom/nodes/attributes.html?label=experimental&label=master&aligned&q=nodes%2Fattributes

polyfills/Element/prototype/getAttributeNames/config.toml
romainmenke: Switched ios safari en desktop safari.
Thank you for double checking!

JakeChampion (approved):

Issue #1017, opened 2021-03-16
Polyfill for Object.fromEntries doesn't work for arrays


JiahuanWu:
[Bug Report]

## What

Polyfill for `Object.fromEntries` only works for maps, but not arrays in IE11.

## Details

Example:
```javascript
const arrayTest = [['foo', 'bar'], ['baz', 42]];

// Gets an error: [object, undefined] is not a function
const objectTest = Object.fromEntries(arrayTest);
// Works
const objectTest1 = Object.fromEntries(new Map(arrayTest));
```
Browser: IE11
OS: Windows 10 on virtualBox VM

JakeChampion:
This is not a bug in the polyfill as Object.fromEntries does not work for Arrays in native implementations either.

Here is the test from the specification conformance test suite which shows Object.fromEntries does not support Arrays -- https://github.com/tc39/test262/blob/42bf3a9f7aed3790dc1d829f290dee033df41c5b/test/built-ins/Object/fromEntries/string-entry-primitive-throws.js#L11-L13

zloirock:
@JakeChampion it's another case.

JakeChampion:
@zloirock Is it this case? https://github.com/Financial-Times/polyfill-library/blob/e4c9bedf49ca9712d8450dc5297e1e4901b8c007/polyfills/Object/fromEntries/tests.js#L68-L83

zloirock:
@JakeChampion yep.

JakeChampion:
We support this case as well, which the test shows. @JiahuanWu, does your polyfill.io include `Object.fromEntries` in the `features` query parameter?

JiahuanWu:
@JakeChampion Yes, I included `Object.fromEntries` in the `features` query parameter along with the others, but only this one didn't work.   
I changed to use babel for js features. So if it can't be reproduced in environment other than my virtualBox, then I'll close the issue.

JakeChampion:
@JiahuanWu Would you be able to create a small reproducible example for this?

We have tests for Object.fromEntries which run on IE11 and all pass, which makes me think your scenario has some other changes which we are not yet aware of that affect the polyfill

JiahuanWu:
@JakeChampion Sorry for late response. My data is just as simple as the test. And I tried on Parallels, it works fine. So I suppose it's a problem of VirtualBox.   
Thanks for your time. I'll close the issue.

PR #1018, opened 2021-03-16
Add test to ensure Object.fromEntries works for the case described in #1017

JakeChampion:


notlee:
I'm assuming the intention isn't to merge this so I removed the team as reviewers, let me know if that's incorrect :)

PR #1038, opened 2021-04-19
Add polyfill for document.elementsFromPoint

alenaksu:
This PR adds the polyfill for `document.elementsFromPoint`. 

The implementation is based on `document.elementFromPoint` and the `pointer-events` css property for excluding elements from hit testing.

For IE8/9 it uses `visibility` as they don't support `pointer-events` for HTML elements, only for SVGs.


polyfills/document/elementsFromPoint/config.toml
```diff
@@ -0,0 +1,67 @@
...
+# For example, the alias `es5` is used to replicate an ECMAScript 5 environment.
+aliases = []
+
+#### Dependencies
+# List any web platform features that you use in your polyfill.
+# For example:
+# dependencies = [
+#    "Object.defineProperty"
+# ]
+dependencies = []
```
JakeChampion, line 12: The polyfill uses Array.prototype.map on line 24 https://github.com/Financial-Times/polyfill-library/pull/1038/files#diff-0726f42f831bc5f6492c044eeffddf5277c06df20b11b41c5c47581740214c51R24

We need to include that as a dependency for browsers which don't have the feature
```suggestion
dependencies = [
	"Array.prototype.map"
]
```

alenaksu: added 86efe2b1656a1e0fea3e93a60106118272311265

polyfills/document/elementsFromPoint/tests.js
```diff
@@ -0,0 +1,54 @@
...
+
+        p.innerText = 'Some text';
+
+        var center = getElementCenter(p);
+
+        var elements = document.elementsFromPoint(center[0], center[1]);
+        elements = [].map.call(elements, function(e) {
+            return e.tagName.toLowerCase();
+        });
+        proclaim.deepStrictEqual(elements, ['p', 'div', 'body', 'html']);
```
JakeChampion, line 27: We can't use Array.prototype.map in the tests because IE 8 does not support `Array.prototype.map`.

We can replace it with a for-loop like below:
```suggestion
        var elements = document.elementsFromPoint(center[0], center[1]);
        var elementTagNames = [];
        for (var i = 0; i < elements.length; I++) {
        	elementTagNames.push(element[i].tagName.toLowerCase());
        }
        proclaim.deepStrictEqual(elements, ['p', 'div', 'body', 'html']);
```

JakeChampion:
@alenaksu the tests for this are failing in nearly all of the browsers, would you be able to look into fixing the tests?

alenaksu:
> @alenaksu the tests for this are failing in nearly all of the browsers, would you be able to look into fixing the tests?

Yes, of course. I'm looking into it. I'll push a fix as soon as possible

alenaksu:
There are still some issues with IE8 and IE9, I'm updating the polyfill and making some tests

alenaksu:
hey @JakeChampion, I see that IE8 is still failing even though I tested it manually several times on Windows XP and it worked. is there any chance to get more informations on why it is failing?

JakeChampion:
Hi @alenaksu, the only information I have is from the test result, which is below:
```
Failures:
 - ie/8.0:
    -> document.elementsFromPoint document.elementsFromPoint returns all the elements at the specified coordinates
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=document.elementsFromPoint&feature=document.elementsFromPoint
       p deepStrictEqual p,div,body,html
```

What other information would you like to have seen? Maybe I can find it

romainmenke:
It is failing because CI runs the test in an iFrame.

Compare locally :

- with iframe : http://bs-local.com:9876/?includePolyfills=yes&always=yes&feature=document.elementsFromPoint
- http://bs-local.com:9876/test?includePolyfills=yes&always=yes&feature=document.elementsFromPoint

No idea why this changes the behaviour.

polyfills/document/elementsFromPoint/polyfill.js
```diff
@@ -0,0 +1,63 @@
+document.elementsFromPoint = document.msElementsFromPoint || function elementsFromPoint(x, y) {
+    var stack = [];
+    var element = document.elementFromPoint(x, y);
```
romainmenke, line 3:
```suggestion
document.elementsFromPoint = function elementsFromPoint(x, y) {
    if ('msElementsFromPoint' in this) {
        return this.msElementsFromPoint(x, y) || [];
    }

    var stack = [];
    var element = document.elementFromPoint(x, y);
```

`msElementsFromPoint` returns `null` instead of an empty array

romainmenke: This can also be tested for in `detect.js`

```
(function () {
	if (!('document' in self && 'elementsFromPoint' in self.document)) {
		return false;
	}

	try {
		var res = document.elementsFromPoint(-9007199254740991, 9007199254740991);
		if (!res) {
			return false;
		}
	} catch (_) {
		return false;
	}

	return true;
})()
```

alenaksu:
> It is failing because CI runs the test in an iFrame.
> 
> Compare locally :
> 
> * with iframe : http://bs-local.com:9876/?includePolyfills=yes&always=yes&feature=document.elementsFromPoint
> * http://bs-local.com:9876/test?includePolyfills=yes&always=yes&feature=document.elementsFromPoint
> 
> No idea why this changes the behaviour.

@JakeChampion @romainmenke 
Thank you, the iframe thing helped me to figure out the problem. It turns out that forcing a reflow (just on IE8) should fix the issue. Actually elementFromPoint should already do that, but for some reason it doesn't into the iframe.

I'm  going to push a commit soon

polyfills/document/elementsFromPoint/polyfill.js
alenaksu: I 've integrated your suggestion, but not the latter as it is quite and edge case and should be happening only on IE where is handled by the first.

JakeChampion (approved):

PR #1041, opened 2021-04-30
Todd gandee/issue 2604/fix intl datetimeformat

tgandee79:
**[Intl.DateTimeFormat  -  Issue 2604](https://github.com/Financial-Times/polyfill-service/issues/2604)**

Timezone Data is not in the Polyfill

Background information
I am using the Intl.DateTimeFormat to label and format with a IANA timezone. I was getting `RangeError` in Bugsnag and it was coming from Safari 10. I thought it was covered in my Polyfill package, but it wasn't

Implementation summary
Added the `all-tz-data.js` to the config path
Added a test to cover the timezone label, it's the easiest test for timezone.

Before & after
Before testing in Safari
image

After just adding the test
image

After adding the path in the config and building
image

### DIFF
--- a/polyfills/Intl/DateTimeFormat/config.toml
+++ b/polyfills/Intl/DateTimeFormat/config.toml
@@ -44,7 +44,10 @@ samsung_mob = "*"
 
 [install]
 module = "@formatjs/intl-datetimeformat"
-paths = [ "polyfill.umd.js" ]
+paths = [
+  "polyfill.umd.js",
+  "add-all-tz.js"
+]
 postinstall = "update.task.js"
 clean = [
   "~locale",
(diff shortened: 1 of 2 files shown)


JakeChampion:
Hi @tgandee79

Did you try loading the specific locale data as a separate feature/polyfill? That's the way we implemented the Intl polyfills in this project.

>Locales must be specified separately by prefixing the locale name with `Intl.DateTimeFormat.~locale`, eg `Intl.DateTimeFormat.~locale.en-GB`.

The reason we do not include the locale data inside the Intl polyfills is because the most applications do not need every single localisation. Loading all the locale data would increase the bundle size to over a megabyte of javascript.

Here is an example of using polyfill.io and loading a specific locale -- `https://polyfill.io/v3/polyfill.js?features=Intl.DateTimeFormat,Intl.DateTimeFormat.~locale-en-UK`

tgandee79:
@JakeChampion - I will implement that change. WIth a quick test on our implementation, I'm seeing the timezone names and locale objects are present.  Thank you.  I'll close my PR and put a comment on the Issue.  Thank you for you time.

PR #1043, opened 2021-05-02
polyfill : HTMLInputElement.prototype.valueAsDate

romainmenke:
Now that [Safari shipped](https://bugs.webkit.org/show_bug.cgi?id=119175) `<input type"date">` it would help to have a polyfill for `valueAsDate`.
This property is useful to get a `Date` object as converting strings to dates is error prone (I personally am bad at doing this correctly each time).

For browsers without native support (no calendar UI) it can still be used with `<input type="date" pattern="...">` to get visitors to enter values in the correct format.

All the date logic seems to hold up against WPT tests and extra edge cases.
Maybe someone with a good understanding of `Date` could take a second look?

------

_fun fact : Old Safari (< 5.1) had native support for `valueAsDate` but it was altered to always throw errors, as there was no `type="date"` support._


polyfills/HTMLInputElement/prototype/valueAsDate/config.toml
```diff
@@ -0,0 +1,39 @@
+#### Aliases
+aliases = []
+
+#### Dependencies
+dependencies = [
+	"Object.defineProperty"
```
JakeChampion, line 6: On line 11 of the polyfill there is a called to Object.getOwnPropertyDescriptor - https://github.com/Financial-Times/polyfill-library/pull/1043/files#diff-dec27411bbff16e8942db416d17c042db40eabaa178872c3fd5a87767370db95R11

We should declare it as a dependency for browser versions like Firefox 35 and lower which do not have it natively
```suggestion
	"Object.defineProperty",
	"Object.getOwnPropertyDescriptor"
```

JakeChampion:
Nice work as usual @romainmenke, thank you for this 😄

JakeChampion (approved):

Issue #1048, opened 2021-05-07
URLSearchParams: keys is not a function


DomenicoCasillo:
Bug Report

## What
URLSearchParams polyfill (v3) seems to be _incomplete_  in Firefox 37 Desktop.

## Details
```javascript
const urlParams = new URLSearchParams(location.search);
console.log(urlParams.keys());
```
The urlParams object is correctly created but on the second line, an error is triggered with
`TypeError: urlParams.keys is not a function`

Browser: Firefox 37 Desktop (Mozilla/5.0 (Windows NT 6.1; rv:37.0) Gecko/20100101 Firefox/37.0)
Device: Desktop
Polyfill URL: https://polyfill.io/v3/polyfill.min.js?features=Promise.prototype.finally%2CPromise%2CURLSearchParams%2CArray.prototype.flatMap

JakeChampion:
@DomenicoCasillo, thanks for the issue, could you please post the full polyfill.io url you are using and the full user-agent of the browser which is seeing the issue please? With that information we should hopefully be able to find the cause of the problem

DomenicoCasillo:
> @DomenicoCasillo, thanks for the issue, could you please post the full polyfill.io url you are using and the full user-agent of the browser which is seeing the issue please? With that information we should hopefully be able to find the cause of the problem

I've updated the initial post with additional details. I can add that even `entries` and `values` are not available.

![image](https://user-images.githubusercontent.com/9551125/117433366-b3354780-af2b-11eb-90df-aa2b44b84553.png)

JakeChampion:
Thanks @DomenicoCasillo - I have confirmed the issue now 👍 

This is happening because the polyfill has not been applied, you can see that if you run `URLSearchParams.toString()`, it will return this:
```
"function URLSearchParams() {
    [native code]
}"
```

That is happening because Firefox 37 passes all the checks in this section of code:
https://github.com/Financial-Times/polyfill-library/blob/2cfa74e454bf6736234bad9bdeca5d96d451ebc2/polyfills/URL/polyfill.js#L36-L45

We need to update that section of code to check for the methods you've mentioned: `keys`, `values`, `entries`.

If you submit a pull-request which does that, I'm happy to review it 👍

romainmenke:
Bit of a rabbit hole :)

One of the reasons it is not working as expected is because there are a few missing dependencies and `Array.from` also seems to be needed in more browser versions than currently stated.

I'll take a look and patch the needed pieces.
[...]

PR #1050, opened 2021-05-07
URLSearchParams : [err] keys is not a function

romainmenke:
fixes : #1048

-----------

As far as I can tell there where 2 issues.
- the detect script was too broad and didn't target new features (e.g. `URLSearchParams.prototype.sort`)
- `Array.from` was used but with it's own fallback and without declaring it as a dependency. Seems to work better with the `Array.from` polyfill.

Some WPT tests included here failed, but nothing truly major imho.

----------

_CI doesn't run on Firefox 37_

test url : http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=URL
browserstack url : https://live.browserstack.com/dashboard#os=OS+X&os_version=El+Capitan&browser=Firefox&browser_version=37.0&zoom_to_fit=true&full_screen=true&resolution=responsive-mode&speed=1

Test `URLSearchParams iterable methods` covers the issue.
This test was previously commented out. 

Screenshot 2021-05-07 at 22 26 35

### DIFF
--- a/polyfills/URL/config.toml
+++ b/polyfills/URL/config.toml
@@ -10,9 +10,10 @@ aliases = [
 dependencies = [
   "Object.defineProperties",
   "Array.prototype.forEach",
-  "Array.isArray"
+  "Array.isArray",
+  "Array.from"
 ]
-notes = [ "Polyfill requires Object getters so fails in IE < 8" ]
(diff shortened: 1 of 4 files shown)


JakeChampion (approved):
Nice work @romainmenke  🎉

PR #1058, opened 2021-05-29
Serve `URL.prototype.toJSON` to Safari <12 and Edge <18.17134

romainmenke:
`URL` polyfill is loaded in Safari `<12` and Edge `<18.17134`.
This erases native support for `URL.prototype.toJSON`.

Matching config for should resolve this.

Screenshot 2021-05-29 at 10 34 28

https://mrhenry.github.io/web-tests/#59933cd3-48cc-4f36-9c52-9ac7b46e5329


romainmenke:
iOS also has unexpected results.
I haven't had the time yet to look into it.

Screenshot 2021-06-08 at 18 06 49

romainmenke:
@JakeChampion Manually verified that iOS 12 and 13 would work after the config change.
These now also receive the `URL` polyfill, but not `URL.prototype.toJSON`

If I am not mistaken they receive the polyfill even when config says `<12` because of the user agent parser.

romainmenke:
Flaky test run in Edge I think.

romainmenke:
@JakeChampion could you rerun CI on this one?
It should be ready to ship :)

Thank you

JakeChampion:
@romainmenke running it now 👍

JakeChampion:
@romainmenke I'm not sure it is a flaky test, I've rerun it several times now and it is always the same browser version and test which fails

romainmenke:
I think I know why it's failing.
The tests run against Edge 18 based on `test/polyfills/browserstackBrowsers.toml`

```toml
[[browsers]]
os = "Windows"
os_version = "10"
browser = "edge"
browser_version = "18.0"
```

But the polyfill config has `edge = "<18.17134"`.
The browser from browserstack has a version higher than `18.17134`.

This is causing issues.

Thank you for running those tests. I assumed it was flaky because it was a seemingly unrelated polyfill.

I'll get back to you with a fix.

romainmenke:
That should do it.

If the combination of the test UA and the `feature` query params results in an empty list we now fallback to the full list for the current UA.

It is never wrong to run the full list for a UA as running a subset is just an optimization to make CI faster.

This run now will run all tests anyway because `test/polyfills/server.js` was changed.
You can test this locally with `http://bs-local.com:9876/?includePolyfills=yes&always=no&feature=URL.prototype.toJSON` against Edge 18 in browserstack.

JakeChampion:
@romainmenke running on ci now 👍

romainmenke:
All green now :)

JakeChampion (approved):

romainmenke:
Thank you @JakeChampion! 🎉 

I noticed that the last version wasn't released yet on npm.
Maybe these latest changes can be released as well?

Issue #870, opened 2020-09-13
URLSearchParams silently fails for FormData under IE 11


wak45:
## Details
<!--
- calling URLSearchParams, passing FormData: body: new URLSearchParams(new FormData(form)),

Code example:
      	fetch("/getdata.php", {
      		method: "POST",
	    	headers: {
    			"Content-Type": "application/x-www-form-urlencoded; charset=UTF-8",
  			},
			body: new URLSearchParams(new FormData(form)),
      	})

<!--
## Additional information

Problem is on the server side, I get php errors:

[13-Sep-2020 12:43:35 UTC] PHP Notice:  Undefined index: name in /getdata.php on line 4
[13-Sep-2020 12:43:35 UTC] PHP Notice:  Undefined index: address in /getdata.php on line 5

As far as I can check, the  URLSearchParams silently fails when passing FormData fails. 

Everything works fine on Chrome and Edge.

lorenz:
In general using fetch with a URLSearchParams body produces an empty request on IE10 and 11. Both APIs are polyfilled there, so there seems to be an issue with them not interacting properly.

romainmenke:
@lorenz Can you provide some code to reproduce and the polyfill.io url?

It doesn't have to be a real fetch request to a real endpoint.

romainmenke:
https://github.com/github/fetch/blob/master/fetch.js#L8

Seems to be a loading order issue.
If I make `URL` a dependency on `fetch` it seems to work.

I don't think we wan't to do that, but maybe we can order the polyfills better?

<img width="413" alt="Screenshot 2021-07-17 at 20 12 11" src="https://user-images.githubusercontent.com/11521496/126046136-87e333b2-b4a0-441b-8098-7ab8b92abc6c.png">

tested with :

```js
if ('Request' in self) {
	it('works with fetch', function (done) {
		// https://github.com/Financial-Times/polyfill-library/issues/870
		new Request('#', { body: new URLSearchParams({ foo: 'baz' }), method: 'POST' }).text().then(function(x) {
			proclaim.equal(x, 'foo=baz');
			done();
		});
	});
}
```

In : `polyfills/URL/tests.js`

`http://bs-local.com:9876/test?includePolyfills=yes&always=no&polyfillCombinations=yes&feature=URL`

lorenz:
My code is quite trivial, your test is already pretty much the same. Nice that you could already determine what's causing the problem, an ordering issue would indeed make sense.

[...]

PR #1068, opened 2021-07-19
fix `URL` + `fetch` interop

romainmenke:
should fix : https://github.com/Financial-Times/polyfill-library/issues/870

There is also a minification issue with an `iife` in the URL polyfill.
I added a few more semicolons around the `iife`'s here.
I can only reproduce it against polyfill.io so I am not sure it is fully fixed.

### DIFF
--- a/polyfills/URL/config.toml
+++ b/polyfills/URL/config.toml
@@ -20,7 +20,7 @@ repo = "https://github.com/inexorabletash/polyfill"
 docs = "https://developer.mozilla.org/en-US/docs/Web/API/URL"
 
 [browsers]
-chrome = "<60"
+chrome = "<61"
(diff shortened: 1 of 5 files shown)


polyfills/URL/config.toml
```diff
@@ -20,7 +20,7 @@ repo = "https://github.com/inexorabletash/polyfill"
 docs = "https://developer.mozilla.org/en-US/docs/Web/API/URL"
 
 [browsers]
-chrome = "<60"
+chrome = "<61"
```
romainmenke, line 23: Chrome 60 still doesn't support `URLSearchParams` constructor with a `Record`

JakeChampion (approved):

PR #1069, opened 2021-07-19
Make MediaQueryList implement the EventTarget interface

santialbo:
According to the documentation (https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList), MediaQueryList should implement the EventTarget interface and the existing methods `addListener` and `removeListener` are deprecated.

This PR adds a very straightforward implementation based on the existing methods.

### DIFF
--- a/polyfills/matchMedia/polyfill.js
+++ b/polyfills/matchMedia/polyfill.js
@@ -53,12 +53,34 @@
 		}
 	};
 
+	MediaQueryList.prototype.addEventListener = function addEventListener(type, listener) {
+		if (type === 'change') {
+			this.addListener(listener);
+		}
+	};
+
 	MediaQueryList.prototype.removeListener = function removeListener(listener) {
 		var listenerIndex = this.listeners.indexOf(listener);
 
 		if (listenerIndex >= 0) {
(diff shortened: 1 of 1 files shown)


romainmenke:
I think this also requires a config change to make sure that we serve the needed polyfill to all browsers : https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#browser_compatibility

santialbo:
> I think this also requires a config change to make sure that we serve the needed polyfill to all browsers : https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#browser_compatibility

I'm not sure I understand why it would need a config change. Do you mean that you would also force this polyfill on browsers that are not implementing the EventTarget interface?

romainmenke:
Maybe we can add this as a separate feature?

If we polyfill these features with the current config it would be confusing why these methods work in older browsers but not in some more recent.

santialbo:
> Maybe we can add this as a separate feature?
> 
> If we polyfill these features with the current config it would be confusing why these methods work in older browsers but not in some more recent.

Makes sense, I'll create another PR

PR #1071, opened 2021-07-20
Add polyfill for making MediaQueryList extend the EventTarget interface

santialbo:
As discussed in #1069. I separated the implementation as a new polyfill so it can be included independently.
I have several concerns:
- I dind't include the implementation of dispatchEvent. If the implementation is not the one in the matchMedia polyfill there's is no way that I know of accessing the list of listeners.
- I'm not sure the path I chose is the correct. Please let me know if you would put it somewhere else and I'll make the change right away.
- I'm not sure I've got the config.toml right.


romainmenke:
see : https://github.com/santialbo/polyfill-library/pull/1

@JakeChampion To really test event listeners on `MediaQueryList` I needed to have iframes that are polyfilled.
To achieve this I extended `test/polyfills`.

In short :

- new endpoint in the test server : `iframe.html`
- uses `createEndpoint` so doesn't add more complexity in `server.js`
- `test-iframe.handlebars` mirrors `test-runner.handlebars` but only loads the polyfill javascript and nothing else
- `test-runner.handlebars` has a `<meta>` tag to expose the correct URL for the URL to tests.

in tests :

```js
var iframe = document.createElement("iframe");
iframe.src = document.getElementById('test-iframe-src').getAttribute('content');
```

The result is a same origin iframe with all polyfills applied.

Feel free to shoot this down :)
Maybe there is another way to test this polyfill without modifying the test server?

JakeChampion:
I don't see another way, I like this approach 👍

santialbo:
This is now updated with your changes @romainmenke

JakeChampion:
The current changes in this pull-request would likely be breaking changes due to the removal of two feature names:
- `MediaQueryList.prototype.EventTarget`
- `MediaQueryList.prototype.dispatchEvent`

If we wanted to keep this change backwards compatible, we should add those two feature names as aliases to the file `polyfills/MediaQueryList/prototype/addEventListener/config.toml`

romainmenke:
@JakeChampion Is it possible you were looking at `New changes since you last viewed` ?
I removed those two aliases from this PR because the features are not implemented (or not fully).
They were however new additions, not existing aliases on the main branch.

JakeChampion:
> @JakeChampion Is it possible you were looking at `New changes since you last viewed` ?
> I removed those two aliases from this PR because the features are not implemented (or not fully).
> They were however new additions, not existing aliases on the main branch.

Yep 🤦 - good work noticing what I did there 😅

JakeChampion (approved):

PR #1076, opened 2021-07-21
add timeZone data in all browsers that need Intl.DateTimeFormat

romainmenke:
fixes : #1075

Exact copy of `Intl.DateTimeFormat` browser config.


JakeChampion:
@romainmenke Do these two browser configs need to stay in sync? If so, let's add a comment above both of them to explain this, hopefully that helps ensure they don't end up out of sync 🤞

romainmenke:
They need to stay in sync with the browser compatibility list from `Intl.DateTimeFormat`.
I added comments, I hope those are clear :)

JakeChampion (approved):

JakeChampion:
Thanks as usual for your contributions @romainmenke 🙇

PR #1077, opened 2021-07-25
check for AbortSignal feature in fetch detect.js

romainmenke:
The `fetch` polyfill `detect.js` only checked against basic support.

This change will make more feature available even when using the `gated` flag.

### DIFF
--- a/polyfills/fetch/config.toml
+++ b/polyfills/fetch/config.toml
@@ -1,5 +1,6 @@
 dependencies = [
 	"Array.prototype.forEach",
+	"Array.isArray",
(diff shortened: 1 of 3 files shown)


polyfills/fetch/config.toml
```diff
@@ -1,5 +1,6 @@
 dependencies = [
 	"Array.prototype.forEach",
+	"Array.isArray",
```
romainmenke, line 3: Used in `function Headers(headers)`

JakeChampion (approved):

PR #1078, opened 2021-08-02
Serve Array.prototype.find to Opera Mobile version < 32

matzimowski:
it's a similar case to https://github.com/Financial-Times/polyfill-library/pull/470

basing on https://caniuse.com/?search=Array.prototype.find `Array.prototype.find` is supported on Opera Mobile from v64.

Screenshot 2021-08-02 at 15 22 28

there is no polyfill for i.e. Opera Mobile Version: 34.0.2254 on Android Version: 5.1

user agent:
Mozilla/5.0 (Linux; U; Android 5.1; GN5001 Build/LMY47D; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/43.0.2357.121 Mobile Safari/537.36 OPR/34.0.2254.126584

https://polyfill.io/v3/polyfill.js?features=Array.prototype.find&UA=Mozilla/5.0%20(Linux;%20U;%20Android%205.1;%20GN5001%20Build/LMY47D;%20wv)%20AppleWebKit/537.36%20(KHTML,%20like%20Gecko)%20Version/4.0%20Chrome/43.0.2357.121%20Mobile%20Safari/537.36%20OPR/34.0.2254.126584

returns:
```
/* Polyfill service v3.105.0
 * For detailed credits and licence information see https://github.com/financial-times/polyfill-service.
 * 
 * Features requested: Array.prototype.find
 *  */


/* No polyfills needed for current settings and browser */
```

and a screenshot from sentry.io tool:

Screenshot 2021-08-02 at 15 31 30


romainmenke:
Looks like incomplete caniuse data.
Opera Mobile should have had this in version 33 (blink version 46).

Opera Mobile 64 is just the most recent version.

https://github.com/mdn/browser-compat-data/blob/main/browsers/opera_android.json#L158-L164

matzimowski:
> Looks like incomplete caniuse data.
> Opera Mobile should have had this in version 33 (blink version 46).
> 
> Opera Mobile 64 is just the most recent version.
> 
> https://github.com/mdn/browser-compat-data/blob/main/browsers/opera_android.json#L158-L164
> 
> also unsure why we serve to chrome 45 :
> 
> https://github.com/Financial-Times/polyfill-library/blob/094ecf6d6a27a86ec27a0b124d13a37e3821e65c/polyfills/Array/prototype/find/config.toml#L17
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find
> 
> That would make it Opera Mobile version 32 (blink version 45)

thanks a lot for double-checking it 🙏 

even if they introduced it in v32 (if I get it right) it still fails for us on v34.

let me know if I should apply any adjustments to my change.

romainmenke:
Is it possible that this is Opera Mini and not Opera Mobile?

matzimowski:
> ~Is it possible that this is Opera Mini and not Opera Mobile?~
> 
> Don't think it's Mini, but definitely something weird about that user agent string.
> Will do some more digging later.

hm, i cannot except for 100% it's not Mini.
`Opera Mobile` is just how [sentry](https://sentry.io/) recognises it.

of course, take ur time, thanks for investigating it!

JakeChampion:
Note: The user-agent is normalized to op_mob/34.0.0

```bash
❯ http 'https://polyfill.io/v3/normalizeUa' 'user-agent:  Mozilla/5.0 (Linux; U; Android 5.1; GN5001 Build/LMY47D; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/43.0.2357.121 Mobile Safari/537.36 OPR/34.0.2254.126584'
HTTP/1.1 200 OK
Accept-Ranges: bytes
Connection: close
Content-Length: 13
Content-Type: text/plain; charset=utf-8
Date: Mon, 02 Aug 2021 14:33:51 GMT
Normalized-User-Agent: op_mob/34.0.0
Retry-After: 0
Server-Timing: HIT-SYNTH, fastly;desc="Edge time";dur=0

op_mob/34.0.0
```

romainmenke:
> Note: The user-agent is normalized to op_mob/34.0.0

Same result as our user-agent parser.

`{"browser":{"Opera":34},"operating_system":"Android","browser_engine":{"Blink":43},"bot":null}`

----------

The weird thing is that this combination of browser version and blink engine should not exist.

https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser#Opera_2015
https://help.opera.com/en/opera-version-history/

I also can't find references to Opera Mobile having multiple engine versions for the same browser version.

When checking ua strings in the wild there are multiple, even one with blink 66
https://user-agents.net/download?browser_type=browser&browser=opera-mobile

Makes it hard to set the correct config here :)

matzimowski:
is there any chance it's not opera mobile/mini at all?

https://www.browscap.org/ua-lookup (not sure if we can rely on that) doesn't even mention Opera here for provided user-agent string:
Screenshot 2021-08-02 at 16 48 25

romainmenke:
The user agent string would still detect this as Opera 34 so updating config for Android webview wouldn't do much :/

Adding Opera Mobile to the config based on the blink engine version seems safe to me, but wouldn't help the visitor with that UA.

Also now curious what this user agent is :)

romainmenke:
Gave it some thought and I think we have two options here:

- support the officially released versions and add Opera Mobile <32 to this config of this polyfill.
- do some digging, mapping UA's used in the wild and set the version to the highest Opera Mobile version using Blink <45

I think option 2 would create issues in the long term. Not only this polyfill, but any polyfill would require a config change.
We can also never test this, as tools like Browserstack (in which we can't test Opera Mobile) only support browser official releases.

@JakeChampion thoughts?

JakeChampion:
I agree with your thinking on this @romainmenke. Lets add Opera Mobile < 32 👍

polyfills/Array/prototype/find/config.toml
```diff
@@ -21,6 +21,7 @@ firefox = "<24"
 ie = "*"
 ie_mob = "10 - *"
 opera = "<32"
+op_mob = "<64"
```
romainmenke, line 24:
```suggestion
op_mob = "<32"
```

as discussed here : https://github.com/Financial-Times/polyfill-library/pull/1078#issuecomment-892009972

JakeChampion (approved):

Issue #1082, opened 2021-08-09
Serve IntersectionObserver to iOS 14


JakeChampion:
## What
There are [reports on Twitter](https://twitter.com/sangster/status/1424736454882648069) that iOS 14 sometimes does not have IntersectionObserver - we should perhaps serve the IntersectionObserver to all iOS 14 users to ensure the feature is always available. We could wrap the polyfill in a feature detect to ensure we don't polyfill over the feature if it is already available.

## Details
Reports from Etsy and N26 engineers that sometimes iOS 14 does not have IntersectionObserver enabled
-- https://twitter.com/sangster/status/1424736454882648069

romainmenke:
tricky issue

Would be interesting to know if this is purely caused by the experimental feature flags in browser settings being turned off.

JakeChampion:
Tricky indeed. I don't know if we could answer that question unfortunately. Perhaps @salemhilal could help us try and find that answer? 🤞

salemhilal:
I'm actually going to defer to @mq2thez for that one; he's the one who did the initial research that concluded that it's possible for IntersectionObserver to be disabled in the first place, so he'd likely have more context as to when it might be missing.

mq2thez:
While it's not impossible that that is the case, I think it's pretty unlikely.

We were seeing error rates for `IntersectionObserver` being undefined that were high enough that it felt like it _had_ to be something in the OS doing it -- several hundred errors within the first day of the new experiment being enabled, from a broad range of users on iOS >= 12. We also got external error reports about issues, and when interviewing those users couldn't find any kind of obvious pattern (accessibility settings, iOS content blockers, privacy choices, etc). Most of these users weren't technical, though, so it's possible we missed something when talking to them.

Manually disabling it was one way that I was able to replicate this, but I suspect that the OS can do it in other ways as well. I've tried tweeting at a number of different Safari / Webkit folks with no reply, though.

romainmenke:
Do you have any stats on the percentage of affected visitors?

---------

> we should perhaps serve the IntersectionObserver to all iOS 14 users to ensure the feature is always available. We could wrap the polyfill in a feature detect to ensure we don't polyfill over the feature if it is already available.

I think that is indeed the best way forward.
If at some point we understand the issue better and decide that it shouldn't be fixed after all, we can revert it.

[...]

PR #1083, opened 2021-08-10
Serve IntersectionObserver to all iOS Safari versions

JakeChampion:
There are reports on Twitter (https://twitter.com/sangster/status/1424736454882648069) that iOS 14 sometimes does not have IntersectionObserver - we should serve the IntersectionObserver to all iOS 14 users to ensure the feature is always available. Resolves https://github.com/Financial-Times/polyfill-library/issues/1082


JakeChampion:
Should we inline the feature detect as well? That way we don't override native implementations which would be on some of the iOS browsers --
```js
'IntersectionObserver' in window &&
'IntersectionObserverEntry' in window &&
'intersectionRatio' in window.IntersectionObserverEntry.prototype
```

What do you think about this @romainmenke?

romainmenke:
Sounds good to me!

chee (approved):
I think this is the right choice, because the promise of polyfill.io is that if they've asked for a feature they no longer have to worry about it not working.

Hopefully we can find out what's going on from the webkit team, and maybe follow up with a new PR if that context changes things.

PR #1086, opened 2021-08-15
Do not polyfill WebAnimation on new Safari

bakura10:
WebAnimation is fully supported started from Safari 13.1 on Mac and Safari 13.4 on iOS, so it is not needed to serve this (large) polyfill.


JakeChampion (approved):
Thanks for this @bakura10

Issue #1089, opened 2021-08-21
window.crypto.randomUUID()


elf-pavlik:
Feature request

## What

polyfill.io doesn't provide polyfill for `window.crypto.randomUUID()`

## Details

Spec: https://github.com/WICG/uuid

>## Implementation status
> * [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1197594) - Shipping 92.
> * [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1723674) - Work in progress.
> * [WebKit](https://bugs.webkit.org/show_bug.cgi?id=229240) - Implemented, not shipped yet.

Polyfill: https://github.com/uuidjs/randomUUID

romainmenke:
Have been working on this : https://github.com/romainmenke/polyfill-library/blob/crypto.randomUUID--sincere-snapping-turtle-cbbcdffcf3/polyfills/crypto/randomUUID/polyfill.js

It currently only lacks a good `randomUUID` implementation for IE 8-9 and other browsers lacking `Blob` support.
Haven't had the time yet to compare different approaches.

JakeChampion:
We won't be implementing a polyfill for this, chee has explained the reasoning in a [comment on the draft PR](https://github.com/Financial-Times/polyfill-library/pull/1090#issuecomment-903159185), I have copied it below:

>this seems, to me, like a case where it would be better to fail than to provide a weak implementation. the draft says "Overwrite all elements of array with cryptographically secure random values" which is just not something we can polyfill.

PR #1090, opened 2021-08-21
crypto.randomUUID

romainmenke:
fixes : https://github.com/Financial-Times/polyfill-library/issues/1089

Not entirely sure that the fallback for the oldest browsers is good enough.
Maybe it is in practice as the number of users is lower?

-----------

Could also be best to wait for a second implementation in a major engine.
Currently this is only supported in Chromium


chee:
this seems, to me, like a case where it would be better to fail than to provide a weak implementation. the draft says "Overwrite all elements of array with cryptographically secure random values" which is just not something we can polyfill.

also we don't normally polyfill drafts.

romainmenke:
I agree on failing instead of having a bad implementation. I will update the code and config.

I personally also prefer using the Blob trick for all browsers even if it is a bit slower. Any thoughts on this?

Holding this till the spec settles is definitely the right way to go. I mainly submitted the PR so that the issue has a counter part :)

polyfills/crypto/randomUUID/config.toml
```diff
@@ -0,0 +1,34 @@
...
+bb = "*"
+chrome = "11 - 91"
+edge = "12 - 91"
+firefox = "26 - *"
+ios_saf = "6.1 - *"
+ie = "10 - *"
+ie_mob = "10 - *"
+opera = "15 - *"
+safari = "6.1 - *"
+firefox_mob = "26 - *"
```
romainmenke, line 33:
```suggestion
firefox = "26 - 94"
ios_saf = "6.1 - *"
ie = "10 - *"
ie_mob = "10 - *"
opera = "15 - *"
safari = "6.1 - *"
firefox_mob = "26 - 94"
```

https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/95

polyfills/crypto/randomUUID/config.toml
```diff
@@ -0,0 +1,34 @@
...
+
+#### Dependencies
+dependencies = []
+
+#### Specification Link
+# TODO update spec link when it settles
+spec = "https://wicg.github.io/uuid/"
+
+#### Documentation Link
+docs = "https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues"
```
romainmenke, line 12:
```suggestion
docs = "https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID"
```

PR #1092, opened 2021-08-22
fix typo in RegExp.prototype.flags browsers config

romainmenke:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/flags

Screenshot 2021-08-22 at 09 50 13


JakeChampion (approved):
Nice spot 🙈

PR #1093, opened 2021-08-22
DOMRect.fromRect static method + DOMRect config fixes

romainmenke:


polyfills/DOMRect/config.toml
```diff
@@ -12,6 +12,10 @@ chrome = "<61"
 ios_chr = "<10.1"
 safari = "<10.1"
 ios_saf = "<10.1"
-firefox = "<=30"
-android = "<6"
```
romainmenke, line 16: 99% sure that `6` was a typo and `61` is the correct config.
https://developer.mozilla.org/en-US/docs/Web/API/DOMRect

JakeChampion (approved):

Issue #221, opened 2019-06-19
Feature request: <details> element polyfill


JakeChampion:
## What
The `<details>` element is in all browser except for IE and Edge, let's add a polyfill for it to this service.

## Details

We can look into using a third-party polyfill such as:
- https://github.com/javan/details-element-polyfill
- https://github.com/rstacruz/details-polyfill

We would need to add our own tests for the feature to ensure that whichever polyfill we use is working as expected.

If anyone would like to take up this feature request, feel free to comment on this issue and I can provide some guidance on how to contribute the feature to this project :-)

## Additional info

MDN documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details
Browser compatibility: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Browser_compatibility
Specification: https://html.spec.whatwg.org/multipage/interactive-elements.html#the-details-element

PR #1102, opened 2021-09-23
Details element polyfill

rmcwhae:
Adds a polyfill for HTML `<details>` elements (Issue https://github.com/Financial-Times/polyfill-library/issues/221).

The polyfill is taken from https://github.com/javan/details-element-polyfill. Tests here are ported versions of those found at https://github.com/javan/details-element-polyfill/blob/main/test/index.js.

Question: I pasted https://github.com/javan/details-element-polyfill/blob/main/dist/details-element-polyfill.js into `polyfill.js` here—is this the correct way of including said polyfill? The other approach, seen in some of the other polyfills, would be to `npm i details-element-polyfill` and add the following lines to `config.toml` (not sure if I should then remove the content in `polyfill.js` to avoid having two copies of the polyfill code):
```
[install]
module = "details-element-polyfill"
paths = [ "dist/details-element-polyfill.js" ]
```


JakeChampion:
Hi @rmcwhae - thanks for the contribution!

> Question: I pasted https://github.com/javan/details-element-polyfill/blob/main/dist/details-element-polyfill.js into polyfill.js here—is this the correct way of including said polyfill? The other approach, seen in some of the other polyfills, would be to npm i details-element-polyfill and add the following lines to config.toml (not sure if I should then remove the content in polyfill.js to avoid having two copies of the polyfill code)

If the code pasted into the polyfill.js file in this project is exactly the same as the code in javan/details-element-polyfill then it is preferred to use the `config.toml` approach as we can more easily update to newer versions of the polyfill if any updates are released ☺️

polyfills/HTMLDetailsElement/detect.js
```diff
@@ -0,0 +1 @@
+"HTMLDetailsElement " in self;
```
JakeChampion, line 1: This file is used as a predicate for an if statement, because of that it needs to be an expression, which means it can't have a semicolon at the end

```suggestion
"HTMLDetailsElement " in self
```

polyfills/HTMLDetailsElement/polyfill.js
```diff
@@ -0,0 +1,244 @@
...
+		typeof HTMLDetailsElement != "undefined" &&
+		element instanceof HTMLDetailsElement;
+	var support = {
+		open: "open" in element || elementIsNative,
+		toggle: "ontoggle" in element
+	};
+	var styles =
+		'\ndetails, summary {\n  display: block;\n}\ndetails:not([open]) > *:not(summary) {\n  display: none;\n}\nsummary::before {\n  content: "►";\n  padding-right: 0.3rem;\n  font-size: 0.6rem;\n  cursor: default;\n}\n[open] > summary::before {\n  content: "▼";\n}\n';
+	var _ref = [],
+		forEach = _ref.forEach,
```
JakeChampion, line 21: Array.prototype.forEach will need to be added to the config.toml.dependencies array

polyfills/HTMLDetailsElement/polyfill.js
```diff
@@ -0,0 +1,244 @@
...
+		document.head.insertAdjacentHTML(
+			"afterbegin",
+			"<style>" + styles + "</style>"
+		);
+	}
+	function polyfillProperties() {
+		var prototype = document.createElement("details").constructor.prototype;
+		var setAttribute = prototype.setAttribute,
+			removeAttribute = prototype.removeAttribute;
+		var open = Object.getOwnPropertyDescriptor(prototype, "open");
```
JakeChampion, line 42: Object.getOwnPropertyDescriptor will need to be added to the config.toml.dependencies array

polyfills/HTMLDetailsElement/polyfill.js
```diff
@@ -0,0 +1,244 @@
...
+		Object.defineProperties(prototype, {
```
JakeChampion, line 43: Object.defineProperties will need to be added to the config.toml.dependencies array

polyfills/HTMLDetailsElement/polyfill.js
```diff
@@ -0,0 +1,244 @@
...
+			new MutationObserver(function (mutations) {
+				forEach.call(mutations, function (mutation) {
+					forEach.call(mutation.addedNodes, setAccessibilityAttributes);
+				});
+			}).observe(document.documentElement, {
+				subtree: true,
+				childList: true
+			});
+		} else {
+			document.addEventListener("DOMNodeInserted", function (event) {
```
JakeChampion, line 151: `Event` will need to be added to the config.toml.dependencies array as not all browsers we support have document.addEventListener

polyfills/HTMLDetailsElement/polyfill.js
```diff
@@ -0,0 +1,244 @@
...
+	function findElementsWithTagName(root, tagName) {
+		return (root.tagName == tagName ? [root] : []).concat(
+			typeof root.getElementsByTagName == "function"
+				? slice.call(root.getElementsByTagName(tagName))
+				: []
+		);
+	}
+	function findClosestElementWithTagName(element, tagName) {
+		if (typeof element.closest == "function") {
+			return element.closest(tagName);
```
JakeChampion, line 233: `Element.prototype.closest` could be added to the config.toml.dependencies array instead of using the inlined polyfill in the else block

polyfills/HTMLDetailsElement/polyfill.js
rmcwhae: Updated

polyfills/HTMLDetailsElement/polyfill.js
rmcwhae: Updated

polyfills/HTMLDetailsElement/polyfill.js
rmcwhae: Updated

polyfills/HTMLDetailsElement/polyfill.js
rmcwhae: Updated

polyfills/HTMLDetailsElement/polyfill.js
rmcwhae: Added

rmcwhae:
Thanks for the quick feedback! I have added the dependencies to `config.toml`.

> If the code pasted into the polyfill.js file in this project is exactly the same as the code in javan/details-element-polyfill then it is preferred to use the `config.toml` approach as we can more easily update to newer versions of the polyfill if any updates are released ☺️

Yes the code is exactly the same, so let’s go with `[install]` in `config.toml`. I added `details-element-polyfill` as a devDependency in the root-level package.json. Two more quick questions:
1. Do I include `package-lock.json` in this PR (package installed using Node 12.22.1)
2. Do I keep the pasted code in `polyfill.js`? Now we have it in two places…

JakeChampion:
> * Do I include `package-lock.json` in this PR (package installed using Node 12.22.1)
> 
> * Do I keep the pasted code in `polyfill.js`? Now we have it in two places…

Yes, the `package-lock.json` should be included as it will have changed when installing the new dependency 👍 

You are right, we don't need the `polyfill.js` file now that we have the config.toml.install section defined, which means the file will be created when running `npm run build`. I've removed the file and added it to `.eslintignore` and `.gitignore` 👍

romainmenke:
@rmcwhae Thank you for this!

I went over the code and noticed this bit : 

```js
  function polyfillProperties() {
    var prototype = document.createElement("details").constructor.prototype;
```

The polyfill alters the `HTMLUnknownElement` prototype which will likely cause issues for some users.

```js
	it('does not affect HTMLUnknownElement proto', function () {
		var unknownProto = document.createElement("unknownwefvxcdfg").constructor.prototype;
		proclaim.ok(("undefined" === typeof Object.getOwnPropertyDescriptor(unknownProto, "open")));
	});
```

Screenshot 2021-09-24 at 16 26 12

It also does not expose `HTMLDetailsElement` which is why it fails the feature detect.
This should be fixed upstream.

rmcwhae:
@romainmenke Agreed, hijacking the `HTMLUnknownElement` prototype is not the best practice. Otherwise, this `<details>` polyfill (https://github.com/javan/details-element-polyfill) does seem to offer a more complete implementation than others (e.g. https://github.com/rstacruz/details-polyfill/blob/master/index.js).

I won't have time to look at updating the polyfill library for a bit, so I might put this PR on hold, unless anyone else has any suggestions?

PR #1105, opened 2021-09-28
fix #1103

### LINKED ISSUE #1103: Intl.Listformat not being loaded for macOS Catalina Safari v14.1 and higher
Type: BUG

## What

`Intl.ListFormat` polyfill is not being loaded for Safari versions >= 14.1. I've observed that this feature is not available on macOS Catalina Safari V14.1 whereas it is available on macOS BigSur Safari V14.1

https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Intl/ListFormat/config.toml#L35

## Details

- `Intl.ListFormat` polyfill should be loaded for MacOS Catalina Safari v14.1 and higher
- Try loading this polyfill.io [link](https://polyfill.io/v3/polyfill.js?features=Intl.ListFormat) on macOS Catalina Safari v14.1 or higher. It says that polyfill is not needed for this browser whereas if we try to check the feature availability in the browser console, it's `undefined`.
- macOS version - 10.15.7
- Safari Version - 14.1.2 (15611.3.10.1.5, 15611)

romainmenke:
`Intl.ListFormat` is not supported on macOS Catalina. 
Serving to all Safari versions might be best for now.

see #1103


srikarsam:
Hey @romainmenke, I think we need to update `locale` configs as well?

`polyfills/Intl/ListFormat/~locale`

romainmenke:
@srikarsam Do you mean `Intl.ListFormat.~locale.en` for example?
These configs are generated based on the config for `Intl.ListFormat`.

https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Intl/ListFormat/update.task.js#L38

srikarsam:
Yes. Okay! that's cool! Thank you!

srikarsam:
@romainmenke I am sorry. I just have one last query. By when this fix will be live?

romainmenke:
@srikarsam That takes a while :)
Releases of the library are published every couple of weeks.
It then takes a bit more time before those are published in the polyfill.io hosted service.

In the mean time you can use a lower version (3.104.0 should work)
https://polyfill.io/v3/polyfill.js?features=Intl.ListFormat

JakeChampion:
@romainmenke I've kicked off the tests on CI now 👍

JakeChampion (approved):

PR #1106, opened 2021-09-29
fix document detect.js and add minimal tests

romainmenke:
before : 

- polyfill not included
- still passes feature detect
- fails minimal test

Screenshot 2021-09-29 at 20 15 57

checking for `'Document' in self` resolves this.

### DIFF
--- a/polyfills/document/detect.js
+++ b/polyfills/document/detect.js
@@ -1,1 +1,1 @@
-"document" in self
+"document" in self && "Document" in self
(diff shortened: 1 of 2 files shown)


polyfills/document/detect.js
```diff
@@ -1 +1 @@
-"document" in self
+"document" in self && "Document" in self
```
chee, line 1: :$

chee (approved):

Issue #995, opened 2021-02-19
String.protoype.matchAll missing


s-cork:
## What
missing `String.prototype.matchAll`

## Details
See this issue #159 which reported some missing es2019 features. `String.prototype.matchAll` must have got lost.

JakeChampion:
> See this issue #159 which reported some missing es2019 features. `String.prototype.matchAll` must have got lost.

`String.prototype.matchAll` was added in ECMAScript 2020. ECMAScript 2020 was ratified in June 2020.

Yes, we need to add a polyfill for this.

PR #1107, opened 2021-10-05
Adds `String.prototype.matchAll` polyfill

mhassan1:
Resolves https://github.com/Financial-Times/polyfill-library/issues/995


mhassan1:
I addressed those two comments. I saw some failing tests in the first run. Do we need another run?

PR #1116, opened 2021-11-10
fix(ResizeObserver): correct safari versions per MDN

cdaringe:
# Problem

MDN & caniuse claim different values for resize observer.

ref: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#browser_compatibility

Aside: should we find a slick way to pull in caniuse-lite into populating these toml configs?

# Solution

Update versions


HenrikJoreteg:
@JakeChampion Hey Jake! Just wanted to +1 this thing... turns out ResizeObserver is definitely in Safari these days, would be great to avoid shipping unneeded polyfills:

Screen Shot 2021-11-10 at 3 21 19 PM

The resize observer polyfill is like ~4k and if you happen to be working on a site for a super-sized online retailer like what Chris and I are doing... shaving bytes === earning dollars :)

HenrikJoreteg:
For quick reference: 
Can I Use: https://caniuse.com/?search=resizeobserver
MDN: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#browser_compatibility

JakeChampion (approved):

PR #1119, opened 2021-11-11
Update config.toml for DOMTokenList.prototype.@@iterator

romainmenke:
Apparently `9-12` is an invalid semver range.
`semver.satisfies`  just returns `false` in case of errors, so this went undetected.

https://github.com/Financial-Times/polyfill-useragent-normaliser/blob/master/compilers/node.js#L128
https://github.com/npm/node-semver/blob/main/functions/satisfies.js#L6


chee (approved):
wow

hmm, maybe we should have a test that runs all the versions through `semver.valid`

PR #1120, opened 2021-11-11
add check for valid version ranges in config.toml

romainmenke:
see : https://github.com/Financial-Times/polyfill-library/pull/1119#pullrequestreview-803625104

- invalid ranges in config.toml now fail the build process
- removed old `buildsources.js` (the parallel script has been in use for some time)

Example error : 

```
{
  name: 'Build error',
  message: 'Error while building: DOMTokenList/prototype/@@iterator',
  error: {
    message: 'Polyfill DOMTokenList.prototype.@@iterator has an incorrect version range for ie: 9-12',
    error: {}
  }
}
```


chee (approved):
incredible speed

polyfills/String/prototype/replaceAll/config.toml
```diff
@@ -39,6 +39,10 @@ docs = "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global
 # Third-party code can use any license which is listed on https://opensource.org/licenses
 license = "MIT"
 
+notes = [
```
romainmenke, line 42: This one was interesting.

I got this error :

```
{
  name: 'Build error',
  message: 'Error while building: String/prototype/replaceAll',
  error: {
    message: "Polyfill String.prototype.replaceAll has an incorrect version range for notes: Polyfill for RegExp.prototype[@@replace] is missing. This polyfill won't work fully as expected untill that polyfill has been implemented.",
    error: {}
  }
}
```

Indicating that `notes` was seen as part of the browsers list.

PR #1122, opened 2021-11-11
allow workflow_dispatch in test-polyfills.yml

romainmenke:
Currently running CI from a fork is no longer possible on the forked repo.
Allowing `workflow_dispatch` makes it possible to run CI on the fork.

It does require bringing your own credentials for browserstack.
But for regular contributors it is very handy to run all tests before opening a pull request.


JakeChampion (approved):

romainmenke:
@JakeChampion Sorry, this doesn't work, in hindsight this is now obvious.
Maybe best to just revert this change.

I'll have to think about this use case.

PR #1125, opened 2021-11-15
add test workflow for forks and document how to use these

romainmenke:
- manually triggered runs only
- simplified as much as possible (doesn't cancel runs when pushing a new commit quickly)
- add a guide on how to set this up


JakeChampion (commented):
At first glance this looks great, I'll give a detailed look at this later on today 👍

JakeChampion (approved):

Issue #300, opened 2019-08-27
Feature Request: Node.isSameNode for Firefox < 48


rokoroku:
## What
`Node.isSameNode()` method

## Details
`Node.isSameNode()` method is a [living standard](https://dom.spec.whatwg.org/#dom-node-issamenode) but the method is not exists on Firefox < 48.0.0 

## Additional information
https://dom.spec.whatwg.org/#dom-node-issamenode
https://developer.mozilla.org/en-US/docs/Web/API/Node/isSameNode#Browser_compatibility

PR #1136, opened 2021-12-03
Adds `Node.prototype.isSameNode` polyfill

mhassan1:
Resolves https://github.com/Financial-Times/polyfill-library/issues/300

### DIFF
--- a/polyfills/Node/prototype/isSameNode/config.toml
+++ b/polyfills/Node/prototype/isSameNode/config.toml
@@ -1,0 +1,28 @@
+aliases = [
+	"dom4",
+	"default-3.4",
+	"default-3.5",
+	"default-3.6",
+	"default",
+]
+dependencies = [
+	"Element",
+]
+docs = "https://developer.mozilla.org/en-US/docs/Web/API/Node/isSameNode"
+
+[browsers]
+chrome = "<1"
+firefox = "<48"
+safari = "<3"
+edge = "<12"
+edge_mob = "<12"
+ie = "<9"
+ie_mob = "<9"
+firefox_mob = "<48"
+opera = "<12.1"
+opera_mini = "<12.1"
+ios_saf = "<1"
+ios_chr = "<1"
+samsung_mob = "<1"
+bb = "*"
+android = "<1"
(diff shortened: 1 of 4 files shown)


JakeChampion:
Thanks @mhassan1 - this is now running on CI over at https://github.com/Financial-Times/polyfill-library/actions/runs/1545824484

romainmenke:
I started on this a while back and eventually abandoned it because I lacked time.
In my version I added a bunch more cases for other types.

https://github.com/romainmenke/polyfill-library/blob/522b4a391792732c7347c711d9ef514d885125b7/polyfills/Node/prototype/isSameNode/polyfill.js

Maybe this isn't needed in which case you can ignore this comment :)
It's too long ago for me to know exactly if these were important and if there were more issues to be resolved.

JakeChampion:
> I started on this a while back and eventually abandoned it because I lacked time. In my version I added a bunch more cases for other types.
> 
> [romainmenke/polyfill-library@`522b4a3`/polyfills/Node/prototype/isSameNode/polyfill.js](https://github.com/romainmenke/polyfill-library/blob/522b4a391792732c7347c711d9ef514d885125b7/polyfills/Node/prototype/isSameNode/polyfill.js)
> 
> Maybe this isn't needed in which case you can ignore this comment :) It's too long ago for me to know exactly if these were important and if there were more issues to be resolved.

It looks like the polyfill you were writing was aiming to work for Internet Explorer 8, which this polyfill currently is not aiming to do.

It would be good to bring in those tests though as they are quite comprehensive 👍

romainmenke:
Maybe easiest if I open a separate PR later to add the extra tests?

JakeChampion:
> Maybe easiest if I open a separate PR later to add the extra tests?

that is good with me ☺️

JakeChampion (approved):

romainmenke:
@JakeChampion Wanted to add the tests and it's starting to come back to me why this was more work than I expected :)

The full WPT test suite is not achievable as not all node types can be patched.
When comparing text nodes for example you get `Invalid pointer` errors.
Going to leave this as is :)

Thank you @mhassan1 for this polyfill!

PR #1138, opened 2021-12-06
Do not error when requesting properties which exist directly on Object.prototype

### LINKED ISSUE #1137: Library throws an error for properties which exist directly on Object.prototype
This was originally reported in https://github.com/Financial-Times/polyfill-service/issues/2652

## What
If requesting any property which exists on directly on Object.prototype such as:
- `'constructor',`
- `'__defineGetter__'`
- `'__defineSetter__'`
- `'hasOwnProperty'`
- `'__lookupGetter__'`
- `'__lookupSetter__'`
- `'isPrototypeOf'`
- `'propertyIsEnumerable'`
- `'toString'`
- `'valueOf'`
- `'__proto__'`
- `'toLocaleString'`

Then the library will throw an error instead of returning a polyfill bundle.

## Details
Running the below code will result in an error being thrown "TypeError: Cannot read property 'has' of undefined".
```js
const polyfillLibrary = require('polyfill-library');
const options = {
	features: {
		'constructor': {},
		'__defineGetter__': {},
		'__defineSetter__': {},
		'hasOwnProperty': {},
		'__lookupGetter__': {},
		'__lookupSetter__': {},
		'isPrototypeOf': {},
		'propertyIsEnumerable': {},
		'toString': {},
		'valueOf': {},
		'__proto__': {},
		'toLocaleString': {},
	},
	uaString: 'ie/9'
};
await polyfillLibrary.getPolyfills(options);
```

JakeChampion:
Resolves #1137


test/unit/lib/index.js
```diff
@@ -2,7 +2,7 @@
 
 "use strict";
 
-const assert = require('proclaim');
+const {assert} = require('chai');
```
JakeChampion, line 5: This was required because proclaim doesn't work correctly for Objects with no prototype.

For example, this should pass but fails with proclaim:
```js
const assert = require('proclaim');

const expected = Object.create(null);
expected.hello = 'world';

const actual = { hello: 'world' };

assert.deepEqual(actual, expected);
```

package.json
```diff
@@ -137,6 +138,7 @@
     "yaml": "^1.1.0"
   },
   "volta": {
-    "node": "12.22.3"
+    "node": "8.17.0",
```
JakeChampion, line 141: I changed this to align with the version we have in the engines field

romainmenke: I am not sure all build and test scripts will work correctly with node8.
It might be that only the published library has node 8 support.

JakeChampion: Yep, I'm going to be testing them all out - as long as the package tests on CI use Node 8 then that's good enough

JakeChampion: Yes, they do not work on node 8, I've pinned local development to node 12 and npm 6

notlee (approved):

PR #1151, opened 2022-01-24
Remove IE 8 workarounds

mhassan1:
This PR removes all workarounds required for IE 8 from polyfills:
* `supportsGetters` & `supportsSetters` checks
* `document.createEvent` vs. `new Event`
* `element.getClientRects()` in `document.elementsFromPoint`
* `arguments.callee` checks
* `unlistenableWindowEvents` in `Event`
* `HTMLPictureElement` detection

It also removes all mentions of IE 8 from documentation and comments.

### DIFF
--- a/polyfills/Event/polyfill.js
+++ b/polyfills/Event/polyfill.js
@@ -1,22 +1,4 @@
 (function () {
-	var unlistenableWindowEvents = {
-		click: 1,
-		dblclick: 1,
-		keyup: 1,
-		keypress: 1,
-		keydown: 1,
-		mousedown: 1,
-		mouseup: 1,
-		mousemove: 1,
-		mouseover: 1,
-		mouseenter: 1,
-		mouseleave: 1,
-		mouseout: 1,
-		storage: 1,
-		storagecommit: 1,
-		textinput: 1
-	};
-
 	// This polyfill depends on availability of `document` so will not run in a worker
 	// However, we asssume there are no browsers with worker support that lack proper
 	// support for `Event` within the worker
@@ -68,10 +50,6 @@
 			element = this,
 			type = arguments[0],
 			listener = arguments[1];
-
-			if (element === window && type in unlistenableWindowEvents) {
-				throw new Error('In IE8 the event: ' + type + ' is not available on the window object. Please see https://github.com/Financial-Times/polyfill-service/issues/317 for more information.');
-			}
 
 			if (!element._events) {
 				element._events = {};
--- a/polyfills/HTMLPictureElement/detect.js
+++ b/polyfills/HTMLPictureElement/detect.js
@@ -1,2 +1,1 @@
-/* Picturefill does not set the HTMLPictureElement global, and the srcset property of <source> does not appear to be accessible to JS in IE8, so in IE8 the only way to detect that the polyfill has applied is to use a non-standard global exposed by the polyfill */
-'HTMLPictureElement' in self || 'picturefill' in self
+'HTMLSourceElement' in self && 'srcset' in self.HTMLSourceElement.prototype
--- a/polyfills/Map/polyfill.js
+++ b/polyfills/Map/polyfill.js
@@ -1,22 +1,5 @@
 /* global CreateIterResultObject, CreateMethodProperty, GetIterator, IsCallable, IteratorClose, IteratorStep, IteratorValue, OrdinaryCreateFromConstructor, SameValueZero, Type, Symbol */
 (function (global) {
-	var supportsGetters = (function () {
-		try {
-			var a = {};
-			Object.defineProperty(a, 't', {
-				configurable: true,
-				enumerable: false,
-				get: function () {
-					return true;
-				},
-				set: undefined
-			});
-			return !!a.t;
-		} catch (e) {
-			return false;
-		}
-	}());
-
 	// Need an internal counter to assign unique IDs to a key map
 	var _uniqueHashId = 0;
 	// Create a unique key name for storing meta data on functions and objects to enable lookups in hash table
@@ -130,16 +113,6 @@
 		// 3. Set map.[[MapData]] to a new empty List.
 		// Polyfill.io - This step was done as part of step two.
 
-		// Some old engines do not support ES5 getters/setters.  Since Map only requires these for the size property, we can fall back to setting the size property statically each time the size of the map changes.
-		if (!supportsGetters) {
-			Object.defineProperty(map, 'size', {
-				configurable: true,
-				enumerable: false,
-				writable: true,
-				value: 0
-			});
-		}
-
 		// 4. If iterable is not present, let iterable be undefined.
 		var iterable = arguments.length > 0 ? arguments[0] : undefined;
 
@@ -199,9 +172,7 @@
 		} catch (e) {
 			// Polyfill.io - For user agents which do not have iteration methods on argument objects or arrays, we can special case those.
 			if (Array.isArray(iterable) ||
-				Object.prototype.toString.call(iterable) === '[object Arguments]' ||
-				// IE 7 & IE 8 return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
-				(!!iterable.callee)) {
+				Object.prototype.toString.call(iterable) === '[object Arguments]') {
 				var index;
 				var length = iterable.length;
 				for (index = 0; index < length; index++) {
@@ -223,19 +194,15 @@
 	});
 
 	// 23.1.2.2 get Map [ @@species ]
-	if (supportsGetters) {
-		Object.defineProperty(Map, Symbol.species, {
-			configurable: true,
-			enumerable: false,
-			get: function () {
-				// 1. Return the this value.
-				return this;
-			},
-			set: undefined
-		});
-	} else {
-		CreateMethodProperty(Map, Symbol.species, Map);
-	}
+	Object.defineProperty(Map, Symbol.species, {
+		configurable: true,
+		enumerable: false,
+		get: function () {
+			// 1. Return the this value.
+			return this;
+		},
+		set: undefined
+	});
 
 	// 23.1.3.1 Map.prototype.clear ( )
 	CreateMethodProperty(Map.prototype, 'clear', function clear() {
@@ -259,9 +226,6 @@
 				M._values[i] = undefMarker;
 			}
 			this._size = 0;
-			if (!supportsGetters) {
-				this.size = this._size;
-			}
 			// 5a. Clear lookup table
 			this._table = {};
 			// 6. Return undefined.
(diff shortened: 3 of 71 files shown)


romainmenke:
This is very thorough, thank you for going over everything and cleaning up IE8 specific bits.

polyfills/Event/polyfill.js
```diff
@@ -70,7 +70,7 @@
 			listener = arguments[1];
 
 			if (element === window && type in unlistenableWindowEvents) {
-				throw new Error('In IE8 the event: ' + type + ' is not available on the window object. Please see https://github.com/Financial-Times/polyfill-service/issues/317 for more information.');
+				throw new Error('The event: ' + type + ' is not available on the window object. Please see https://github.com/Financial-Times/polyfill-service/issues/317 for more information.');
```
JakeChampion, line 73: We can remove all the code related to the `unlistenableWindowEvents` objetc as that is all only for IE8

polyfills/HTMLPictureElement/detect.js
```diff
@@ -1,2 +1,2 @@
-/* Picturefill does not set the HTMLPictureElement global, and the srcset property of <source> does not appear to be accessible to JS in IE8, so in IE8 the only way to detect that the polyfill has applied is to use a non-standard global exposed by the polyfill */
+/* Picturefill does not set the HTMLPictureElement global, and the srcset property of <source> does not always appear to be accessible to JS, so the only way to detect that the polyfill has applied is to use a non-standard global exposed by the polyfill */
 'HTMLPictureElement' in self || 'picturefill' in self
```
JakeChampion, line 2: We can change this detect to be `'HTMLSourceElement' in self &&  'srcset' in self.HTMLSourceElement.prototype` and then remove the comment as it no longer applies

polyfills/Map/polyfill.js
```diff
@@ -200,7 +173,7 @@
 			// Polyfill.io - For user agents which do not have iteration methods on argument objects or arrays, we can special case those.
 			if (Array.isArray(iterable) ||
 				Object.prototype.toString.call(iterable) === '[object Arguments]' ||
-				// IE 7 & IE 8 return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
+				// some engines return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
 				(!!iterable.callee)) {
```
JakeChampion, line 177: What do you think to removing the `|| (!!iterable.callee)` instead?

polyfills/RegExp/prototype/@@matchAll/polyfill.js
```diff
@@ -24,19 +24,6 @@ CreateMethodProperty(RegExp.prototype, Symbol.matchAll, function (string) {
 	var C = SpeciesConstructor(R, RegExp);
 	// 5. Let flags be ? ToString(? Get(R, "flags")).
 	var flags = ToString(Get(R, 'flags'));
-	// IE8 doesn't have RegExp.prototype.flags support
```
JakeChampion, line 27: According to MDN no version of IE has RegExp.prototype.flags -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/flags#browser_compatibility

polyfills/Set/polyfill.js
```diff
@@ -86,7 +59,7 @@
 			// Polyfill.io - For user agents which do not have iteration methods on argument objects or arrays, we can special case those.
 			if (Array.isArray(iterable) ||
 				Object.prototype.toString.call(iterable) === '[object Arguments]' ||
-				// IE 7 & IE 8 return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
+				// some engines return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
 				(!!iterable.callee)) {
```
JakeChampion, line 63: Same here as reported before, I think we should instead remove ths `|| (!!iterable.callee)`

polyfills/String/prototype/matchAll/polyfill.js
```diff
@@ -13,18 +13,11 @@ CreateMethodProperty(String.prototype, 'matchAll', function matchAll(regexp) {
 		if (isRegExp) {
 			// 2.b.i. Let flags be ? Get(regexp, "flags").
 			var flags = Get(regexp, "flags");
-
-			// IE8 doesn't have RegExp.prototype.flags support, it does have RegExp.prototype.global
+			// 2.b.ii. Perform ? RequireObjectCoercible(flags).
+			RequireObjectCoercible(flags)
 			// 2.b.iii. If ? ToString(flags) does not contain "g", throw a TypeError exception.
-			if (!('flags' in RegExp.prototype) && regexp.global !== true) {
```
JakeChampion, line 19: Again, MDN reports that no IE versions have RegExp.prototype.flags, we need to do a test on this to confirm -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/flags#browser_compatibility

polyfills/String/prototype/replaceAll/polyfill.js
```diff
@@ -13,18 +13,11 @@ CreateMethodProperty(String.prototype, 'replaceAll', function replaceAll(searchV
 		if (isRegExp) {
 			// 2.b.i. Let flags be ? Get(searchValue, "flags").
 			var flags = Get(searchValue, "flags");
-
-			// IE8 doesn't have RegExp.prototype.flags support, it does have RegExp.prototype.global
+			// 2.b.ii. Perform ? RequireObjectCoercible(flags).
+			RequireObjectCoercible(flags)
 			// 2.b.iii. If ? ToString(flags) does not contain "g", throw a TypeError exception.
-			if (!('flags' in RegExp.prototype) && searchValue.global !== true) {
```
JakeChampion, line 19: Same here, we need to test that no other IE versions need this -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/flags#browser_compatibility

polyfills/WeakMap/polyfill.js
```diff
@@ -78,7 +78,7 @@
 			// Polyfill.io - For user agents which do not have iteration methods on argument objects or arrays, we can special case those.
 			if (Array.isArray(iterable) ||
 				Object.prototype.toString.call(iterable) === '[object Arguments]' ||
-				// IE 7 & IE 8 return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
+				// some engines return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
 				(!!iterable.callee)) {
```
JakeChampion, line 82: Let's remove this instead `|| (!!iterable.callee)`

polyfills/WeakSet/polyfill.js
```diff
@@ -56,7 +56,7 @@
 			// Polyfill.io - For user agents which do not have iteration methods on argument objects or arrays, we can special case those.
 			if (IsArray(iterable) ||
 				Object.prototype.toString.call(iterable) === '[object Arguments]' ||
-				// IE 7 & IE 8 return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
+				// some engines return '[object Object]' for the arguments object, we can detect by checking for the existence of the callee property
 				(!!iterable.callee)) {
```
JakeChampion, line 60: Let's remove this instead `|| (!!iterable.callee)`

polyfills/document/elementsFromPoint/polyfill.js
```diff
@@ -23,7 +23,7 @@ document.elementsFromPoint = function elementsFromPoint(x, y) {
 			element.style.setProperty(name, value, priority);
 		} else {
 			element.style[name] = value;
-			// need to force a reflow on IE8 in some cases
+			// need to force a reflow in some cases
 			element.getClientRects();
```
JakeChampion, line 27: This line might no longer be required now we don't support ie8

JakeChampion:
Thanks for the pull-request @mhassan1 - I've triggered a test on CI which can be seen here --> https://github.com/Financial-Times/polyfill-library/runs/4929426492?check_suite_focus=true

polyfills/Number/parseFloat/polyfill.js
```diff
@@ -1,18 +1,4 @@
 /* global CreateMethodProperty */
-(function (nativeparseFloat, global) {
-	var parseFloat = function parseFloat(str) {
-		var string = String(str).trim();
-		var result = nativeparseFloat(string);
-		return result === 0 && string.charAt(0) == '-' ? -0 : result;
-	}
```
JakeChampion, line 7: We need to keep this part, it is the actual polyfill for older browsers which have parseFloat but not a fully standards compliant implementation

polyfills/Number/parseInt/polyfill.js
```diff
@@ -1,18 +1,4 @@
 /* global CreateMethodProperty */
-(function (nativeParseInt, global) {
-	// Polyfill.io - IE 8's parseInt is incorrect
-	var parseInt = function parseInt(str, radix) {
-		var string = String(str).trim();
```
JakeChampion, line 5: We need to keep this part, it is the actual polyfill for older browsers which have parseInt but not a fully standards compliant implementation

polyfills/RegExp/prototype/@@matchAll/polyfill.js
mhassan1: Good catch, I will put it back

polyfills/String/prototype/matchAll/polyfill.js
mhassan1: Good catch, I will put it back

polyfills/String/prototype/replaceAll/polyfill.js
mhassan1: Good catch, I will put it back

polyfills/HTMLPictureElement/detect.js
```diff
@@ -1,2 +1 @@
-/* Picturefill does not set the HTMLPictureElement global, and the srcset property of <source> does not appear to be accessible to JS in IE8, so in IE8 the only way to detect that the polyfill has applied is to use a non-standard global exposed by the polyfill */
-'HTMLPictureElement' in self || 'picturefill' in self
+'HTMLSourceElement' in self && 'srcset' in self.HTMLSourceElement.prototype
```
mhassan1, line 1: @JakeChampion Should this be `'HTMLPictureElement' in self`, instead of `'HTMLSourceElement' in self`?

JakeChampion: This should be fine as is 👍

mhassan1:
@JakeChampion This should be ready for re-review and re-test.

JakeChampion:
it all passes 👍  https://github.com/Financial-Times/polyfill-library/actions/runs/1757716346

JakeChampion (approved):

PR #1182, opened 2022-04-14
EventSource polyfill is now protestware

romainmenke:
see : https://github.com/Yaffle/EventSource/commit/de137927e13d8afac153d2485152ccec48948a7a

----

This change aims to protect the polyfill-library from accidentally updating this dependency.
There is a lock file and CI uses `npm ci` so immediate risk is low.

But still safest to commit the polyfill file and remove it from the npm flow.

### DIFF
--- a/polyfills/EventSource/polyfill.js
+++ b/polyfills/EventSource/polyfill.js
@@ -6,6 +6,28 @@ if (!("EventSource"in self&&"function"==typeof self.EventSource
  * eventsource.js
  * Available under MIT License (MIT)
  * https://github.com/Yaffle/EventSource/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2012 vic99999@yandex.ru
+ *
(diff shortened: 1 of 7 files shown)


polyfills/EventSource/polyfill.js
```diff
@@ -0,0 +1,1053 @@
+(function(undefined) {
+if (!("EventSource"in self&&"function"==typeof self.EventSource
+)) {
+// EventSource
+/** @license
+ * eventsource.js
+ * Available under MIT License (MIT)
+ * https://github.com/Yaffle/EventSource/
```
chee, line 8: i think we might need to include the full license here legally? 

> The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.

chee (approved):
let's go. though i did add a comment

polyfills/EventSource/polyfill.js
romainmenke: @chee Thank you for checking this!
The full license has been added.

Issue #1175, opened 2022-03-17
Github actions now support re-running individual jobs.


romainmenke:
https://github.com/Financial-Times/polyfill-library/blob/0ece79ce3238190eded65f3a9432a0ca6bf5d65d/.github/workflows/test-polyfills.yml#L54

https://github.blog/changelog/2022-03-17-github-actions-re-run-only-failed-or-individual-jobs/

This might simplify the workflows.
I haven't tested it yet but will try to find time for this.

PR #1183, opened 2022-04-23
simplify test action as github now supports partial reruns

romainmenke:
fixes : https://github.com/Financial-Times/polyfill-library/issues/1175

_I can not test this action as it won't run from a fork._

Steps to verify that it works as intended :
- start a test run
- cancel a job (either on GitHub or on Browserstack)
- wait for other jobs to finish
- re-run failed jobs only should be available and choosing this should trigger a partial re-run.

### DIFF
--- a/.github/workflows/test-polyfills.yml
+++ b/.github/workflows/test-polyfills.yml
@@ -40,18 +40,9 @@ jobs:
     - run: npm run build
       if: steps.cache-dist.outputs.cache-hit != 'true'
 
-    - run: echo "${{ matrix.browser }}" > $GITHUB_WORKSPACE/witness-${{ matrix.browser }}
-    - name: record ${{ matrix.browser }} witness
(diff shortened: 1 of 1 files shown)


romainmenke:
@chee Are the failures expected here or did I change something inadvertently that is causing these?

chee:
> @chee Are the failures expected here or did I change something inadvertently that is causing these?

i think neither. i tried pushing this commit on here and opening a draft PR with it and it seems to have HATED IT. it can't find the PR branch, also when it was running the tests over there they also showed over here. i think that maybe it is confused about having the same commit on two different remotes? anyway, i don't think it is your fault (i think it is mine and also github's) and i'm happy to merge this in and see what happens.

romainmenke:
Sounds great :)
It can always be reverted if there is anything unexpected.

PR #1185, opened 2022-04-29
Polyfill for HTMLCollection.prototype.@@iterator

plesiecki:
Hello,

Old Edge lacks of `HTMLCollection.prototype.@@iterator`, so here goes the polyfill for it.

Proof
image

### DIFF
--- a/polyfills/HTMLCollection/prototype/@@iterator/config.toml
+++ b/polyfills/HTMLCollection/prototype/@@iterator/config.toml
@@ -3,3 +3,7 @@ dependencies = [ "_ArrayIterator", "Symbol.iterator" ]
 [browsers]
 edge = "<18"
 edge_mob = "<18"
+chrome = "<51"
+firefox = "<13"
+ie = "*"
+safari = "<10"
--- a/polyfills/HTMLCollection/prototype/@@iterator/config.toml
+++ b/polyfills/HTMLCollection/prototype/@@iterator/config.toml
@@ -1,0 +1,5 @@
+dependencies = [ "_ArrayIterator", "Symbol.iterator" ]
+
+[browsers]
+edge = "<18"
+edge_mob = "<18"
--- a/polyfills/HTMLCollection/prototype/@@iterator/detect.js
+++ b/polyfills/HTMLCollection/prototype/@@iterator/detect.js
@@ -1,0 +1,1 @@
+'Symbol' in self && 'iterator' in self.Symbol && self.Symbol.iterator in self.HTMLCollection.prototype && typeof self.HTMLCollection.prototype[self.Symbol.iterator] === "function"
--- a/polyfills/HTMLCollection/prototype/@@iterator/tests.js
+++ b/polyfills/HTMLCollection/prototype/@@iterator/tests.js
@@ -1,12 +1,11 @@
 /* eslint-env mocha, browser */
 /* global proclaim, Symbol */
 
-var collectionId = 0;
 function getHTMLCollection() {
-	collectionId++;
-	document.body.appendChild(document.createElement('something' + collectionId));
-	document.body.appendChild(document.createElement('something' + collectionId));
-	return document.getElementsByTagName('something' + collectionId);
+	var element = document.createElement('div');
+	element.appendChild(document.createElement('span'));
+	element.appendChild(document.createElement('span'));
+	return element.children;
 }
 
 it('exists', function () {
--- a/polyfills/HTMLCollection/prototype/@@iterator/tests.js
+++ b/polyfills/HTMLCollection/prototype/@@iterator/tests.js
@@ -1,11 +1,12 @@
 /* eslint-env mocha, browser */
 /* global proclaim, Symbol */
 
+var collectionId = 0;
 function getHTMLCollection() {
-	var fragment = document.createDocumentFragment();
-	fragment.appendChild(document.createElement('div'));
-	fragment.appendChild(document.createElement('div'));
-	return fragment.children;
+	collectionId++;
+	document.body.appendChild(document.createElement('something' + collectionId));
+	document.body.appendChild(document.createElement('something' + collectionId));
+	return document.getElementsByTagName('something' + collectionId);
 }
 
 it('exists', function () {
--- a/polyfills/HTMLCollection/prototype/@@iterator/tests.js
+++ b/polyfills/HTMLCollection/prototype/@@iterator/tests.js
@@ -1,0 +1,36 @@
+/* eslint-env mocha, browser */
+/* global proclaim, Symbol */
+
+function getHTMLCollection() {
+	var fragment = document.createDocumentFragment();
+	fragment.appendChild(document.createElement('div'));
+	fragment.appendChild(document.createElement('div'));
+	return fragment.children;
+}
+
+it('exists', function () {
+	if (!Symbol || !Symbol.iterator) {
+		proclaim.fail();
+		return;
+	}
+	proclaim.isInstanceOf(getHTMLCollection()[Symbol.iterator], Function);
+});
+
+it('returns a next-able object', function () {
+	var htmlCollection = getHTMLCollection();
+	var iterator = htmlCollection[Symbol.iterator]();
+
+	proclaim.isInstanceOf(iterator.next, Function);
+	proclaim.deepEqual(iterator.next(), {
+		value: htmlCollection[0],
+		done: false
+	});
+});
+
+it('finally returns a done object', function () {
+	var htmlCollection = getHTMLCollection();
+	var iterator = htmlCollection[Symbol.iterator]();
+	iterator.next();
+	iterator.next();
+	proclaim.equal(iterator.next().done, true);
+});
--- a/polyfills/HTMLCollection/prototype/@@iterator/polyfill.js
+++ b/polyfills/HTMLCollection/prototype/@@iterator/polyfill.js
@@ -1,0 +1,4 @@
+/* global Symbol, ArrayIterator*/
+HTMLCollection.prototype[Symbol.iterator] = function () {
+  return new ArrayIterator(this);
+};


plesiecki:
Hello @JakeChampion. Would you mind taking a look please? Thanks

JakeChampion:
@plesiecki I can take a look but I can no longer run the CI tests or merge the pull-request as I no longer have write access to this project.

polyfills/HTMLCollection/prototype/@@iterator/detect.js
```diff
@@ -0,0 +1 @@
+'Symbol' in self && 'iterator' in self.Symbol && self.Symbol.iterator in self.HTMLCollection.prototype
```
JakeChampion, line 1: This should also check to see if `HTMLCollection.prototype[Symbol.iterator]` is a function

plesiecki: Thanks, done.

plesiecki:
@chee @romainmenke May I ask you for review?

romainmenke:
I also don't have write access here :)
But I can run CI for you to make sure everything passes : https://github.com/romainmenke/polyfill-library/actions/runs/2292523646

I was also wondering if you have a reference for which browsers do or do not support this?

plesiecki:
@romainmenke Thanks for the tip about running tests from fork. 👍 
I don't have a reference like MDN or caniuse. The only thing I've got are these screenshots:
Edge v17
image

Edge v18
image

romainmenke:
re-running CI : https://github.com/romainmenke/polyfill-library/actions/runs/2293298928

To figure out which browsers do or do not support a certain feature I created this : https://github.com/mrhenry/web-tests
I won't have much time today to add a test myself, but happy to review a PR.
Otherwise I will add a test later this week.

plesiecki:
@romainmenke I've noticed the mention of WPT in web-tests' README and it has tests for [HTMLCollection iterator](https://github.com/web-platform-tests/wpt/blob/9a88dfd5bf623a8cb326c154b27f1e3494e11022/dom/collections/HTMLCollection-iterator.html), with preview available [here](https://w3c-test.org/dom/collections/HTMLCollection-iterator.html). If so, do you think web-tests needs more tests?

Also, could you please re-run CI on the latest commit 987a72c? 🙏

romainmenke:
CI now running for the right commit : https://github.com/romainmenke/polyfill-library/actions/runs/2294360316

WPT doesn't track older browser versions so it can not be used to figure out which browser versions historically supported a certain feature.

web-tests uses browserstack to go back and test as much versions as possible.
It also can test features in combination with build tools and/or polyfills to give a guarantee that a feature works for a certain setup.

polyfills/HTMLCollection/prototype/@@iterator/tests.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-env mocha, browser */
+/* global proclaim, Symbol */
+
+function getHTMLCollection() {
+	var fragment = document.createDocumentFragment();
+	fragment.appendChild(document.createElement('div'));
+	fragment.appendChild(document.createElement('div'));
+	return fragment.children;
+}
```
romainmenke, line 9:
```suggestion
var collectionId = 0;
function getHTMLCollection() {
	collectionId++;

	document.body.appendChild(document.createElement('something' + collectionId));
	document.body.appendChild(document.createElement('something' + collectionId));
	return document.getElementsByTagName('something' + collectionId);
}
```

`DocumentFragment.prototype.children` was only added in Edge 16 : https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment/children#browser_compatibility

This makes it less suited in the tests here.

polyfills/HTMLCollection/prototype/@@iterator/tests.js
```diff
@@ -2,10 +2,9 @@
 /* global proclaim, Symbol */
 
 function getHTMLCollection() {
-	var fragment = document.createDocumentFragment();
-	fragment.appendChild(document.createElement('div'));
-	fragment.appendChild(document.createElement('div'));
-	return fragment.children;
+	document.body.appendChild(document.createElement('div'));
+	document.body.appendChild(document.createElement('div'));
+	return document.getElementsByTagName('div');
```
romainmenke, line 7: I also tested this :) , but this fails as each call to this function adds two more divs.

This makes the iterator not "done" after two calls to `next()` in the second test.

polyfills/HTMLCollection/prototype/@@iterator/tests.js
plesiecki: Oh cool, I haven't noticed your suggestion while fixing tests. What do you think about my approach?

polyfills/HTMLCollection/prototype/@@iterator/tests.js
plesiecki: Ok, I'm going to apply your suggestion soon.

plesiecki: done

polyfills/HTMLCollection/prototype/@@iterator/config.toml
```diff
@@ -0,0 +1,5 @@
+dependencies = [ "_ArrayIterator", "Symbol.iterator" ]
+
+[browsers]
+edge = "<18"
+edge_mob = "<18"
```
romainmenke, line 5: see : https://mrhenry.github.io/web-tests/#335c187c-f6e4-4e5e-946f-decfe327b764

```toml
edge = "<18"
edge_mob = "<18"
chrome = "<51"
firefox = "<13"
ie = "*"
safari = "<10"
```

Ideally we map as many browsers as possible by their engine to complete this list as good as possible : 

```toml
android = "*"
bb = "*"
chrome = "*"
edge = "*"
edge_mob = "*"
firefox = "*"
ios_chr = "*"
ios_saf = "*"
ie = "*"
ie_mob = "*"
opera = "*"
op_mini = "*"
safari = "*"
firefox_mob = "*"
samsung_mob = "*"
```

plesiecki: Sure, done.

plesiecki:
@romainmenke Thank you for tremendous support. 
@chee Would you mind running remaining workflows please? Thanks

chee:
hey, @plesiecki ! thanks! i have no idea what i'm doing, but i believe the tests should now be running over here: https://github.com/Financial-Times/polyfill-library/actions/runs/2303262365 !!!!

plesiecki:
Welp, there's an issue with tests. `Request failed with status 401 due to Invalid username or password`
Seems like problem with browserstack's credentials(?).

chee:
> Welp, there's an issue with tests. `Request failed with status 401 due to Invalid username or password`
> Seems like problem with browserstack's credentials(?).

ah. it's probably jake's secrets. i'll update it tomorrow.

chee:
> > Welp, there's an issue with tests. `Request failed with status 401 due to Invalid username or password`
> > Seems like problem with browserstack's credentials(?).
> 
> ah. it's probably jake's secrets. i'll update it tomorrow.

OK, let's try again!

https://github.com/Financial-Times/polyfill-library/actions/runs/2303262365

polyfills/HTMLCollection/prototype/@@iterator/tests.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-env mocha, browser */
+/* global proclaim, Symbol */
+
+var collectionId = 0;
+function getHTMLCollection() {
+	collectionId++;
+	document.body.appendChild(document.createElement('something' + collectionId));
+	document.body.appendChild(document.createElement('something' + collectionId));
+	return document.getElementsByTagName('something' + collectionId);
```
plesiecki, line 9: Unfortunately Chrome 29-32 doesn't return an instance of `HTMLCollection` here but a `NodeList`

plesiecki:
Had to try different approach for testing this feature.
@chee Please, could you run test again?

polyfills/HTMLCollection/prototype/@@iterator/tests.js
```diff
@@ -1,12 +1,11 @@
...
 /* global proclaim, Symbol */
 
-var collectionId = 0;
 function getHTMLCollection() {
-	collectionId++;
-	document.body.appendChild(document.createElement('something' + collectionId));
-	document.body.appendChild(document.createElement('something' + collectionId));
-	return document.getElementsByTagName('something' + collectionId);
+	var element = document.createElement('div');
+	elemenet.appendChild(document.createElement('span'));
```
romainmenke, line 6: typo :)

`elemenet` -> `element`

plesiecki: thanks, fixed

plesiecki:
@chee Re-run won't help because it doesn't test the latest version (commit). I guess you should run workflow for that branch again.

chee:
> @chee Re-run won't help because it doesn't test the latest version (commit). I guess you should run workflow for that branch again.

yeah. i'm just updating the branch now.

plesiecki:
All tests passed 🎉 I'm happy to merge and release it 🙏

chee (approved):

plesiecki:
Thanks for the merge! 
Unfortunately [tag-and-release has failed.](https://github.com/Financial-Times/polyfill-library/actions/runs/2333745088). Please help 🙏

chee:
i'm openly weeping. i'll try setting a different secret and running again.

chee:
i've tagged it manually, but it won't exist on polyfill.io until the service is updated. ¿was 4.0 never made live?

romainmenke:
> ¿was 4.0 never made live?

There is always a long period between releasing an update here and deploying it on polyfill.io. A bit of extra safety so that issues can surface before this is pushed to the service.

chee:
> > ¿was 4.0 never made live?
> 
> There is always a long period between releasing an update here and deploying it on polyfill.io. A bit of extra safety so that issues can surface before this is pushed to the service.

ohh! i see. i always thought it was because jake champion was work-shy

JakeChampion:
4.0 drops support for older versions of internet explorer which could cause sites to stop working if we updated the default library version on polyfill.io/v3 (but those sites can/should pin to an older version of polyfill-library).
My thoughts were to make a new polyfill.io/v4 endpoint which defaults the library version to 4.0 but allows previous versions to be selected just like the polyfill.io/v3 endpoint does. We could also internally (in the fastly config) rewrite the polyfill.io/v3 urls to hit polyfill.io/v4 like we do with internally rewriting polyfill.io/v2 urls (the rewrite code is in vcl https://github.com/Financial-Times/polyfill-service/blob/main/fastly/vcl/polyfill-service.vcl#L97-L104 and in js https://github.com/Financial-Times/polyfill-service/blob/main/fastly/c-at-e/src/index.js#L134-L138).

plesiecki:
Very thank you for the release ❤️ Although a release note with PR title would be better rather than a message from the last commit.

chee:
> Very thank you for the release ❤️ Although a release note with PR title would be better rather than a message from the last commit.

thanks that's a really good tip

PR #1187, opened 2022-05-19
Adds `Symbol.toStringTag` to several prototypes

mhassan1:
This PR adds `Symbol.toStringTag` to several prototypes:
- `Map`
- `MapIterator`
- `Set`
- `SetIterator`
- `WeakSet`
- `StringIterator`
- `ArrayIterator`
- `Promise`
- `TypedArray`
- `ArrayBuffer`
- `DataView`

Notes:
- `WeakMap` and `RegExpStringIterator` already had it
- This does not include `JSON`, `Math`, and all DOM objects
- This does not include modifying `Object.prototype.toString` to use `Symbol.toStringTag`


polyfills/Array/prototype/@@iterator/tests.js
```diff
@@ -3,3 +3,9 @@
 it('is an alias to Array.prototype.values', function () {
 	proclaim.deepEqual(Array.prototype[Symbol.iterator], Array.prototype.values);
 });
+
+if (typeof Symbol != 'undefined' && 'toStringTag' in Symbol) {
```
chee, line 7: if it's a dependency, shouldn't this always be `true`? and if it's a dependency wouldn't we want this to fail if it's `undefined` ?

mhassan1: I saw that pattern in a bunch of places ([example](https://github.com/Financial-Times/polyfill-library/blob/871a6c3d2a419a245bfc2c6359990f0beb132da2/polyfills/Array/from/tests.js#L95)), but I agree that it doesn't make sense. I will take it out.

chee: @JakeChampion sorry to disturb you, but would you have any context on this?

JakeChampion: It was because Array.from only makes use of Symbol.iterator but does not itself depend on Symbol.iterator - this meant that it would not polyfill Symbol.iterator when Array.from was polyfilled, so in browsers which had Symbol.iterator (either polyfilled or not) we would test that Array.from would work correct with objects which implemented the Symbol.iterator protocol

chee: https://github.com/Financial-Times/polyfill-library/blob/871a6c3d2a419a245bfc2c6359990f0beb132da2/polyfills/Array/from/config.toml#L15

i guess it started depending on it later?

mhassan1:
@chee @JakeChampion Is anything blocking this PR?

polyfills/ArrayBuffer/prototype/@@toStringTag/config.toml
```diff
@@ -0,0 +1,21 @@
...
+ie = "*"
+ie_mob = "*"
+safari = "<9"
+chrome = "<49"
+firefox = "< 51"
+ios_saf = "< 10"
+opera = "< 37"
+samsung_mob = "<5"
+android = "*"
+bb = "10 - *"
```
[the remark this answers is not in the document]
mhassan1: I used the same versions as the `Symbol.toStringTag` polyfill: https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Symbol/toStringTag/config.toml. Maybe those are also incorrect?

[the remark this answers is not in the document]
romainmenke: 🤔 MDN might have been updated.
Going to run some tests :)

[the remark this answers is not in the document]
romainmenke: Test results are in for `ArrayBuffer` and a bunch of others :
https://mrhenry.github.io/web-tests/#eeb9401d-7a62-48d3-bbe1-b574cace6f98

Looks like MDN is correct (now)

But MDN seems incorrect in other places :
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#browser_compatibility

Here it states that `Map.protototype[@@toStringTag]` is supported before `Symbol.toStringTag` which seems odd.

Tests for `Map.protototype[@@toStringTag]` : https://mrhenry.github.io/web-tests/#decec341-891b-41cc-905f-412c1e8bba23

-----

Things we should update before we can merge this :

- make sure each polyfill has a detect script that also checks for `@@toStringTag`. If we do not have this the `gated` flag will not work as expected
- adjust the support stats in configs so that any browser that needs `@@toStringTag` will receive the polyfill.
- if there is a large gap between main feature support and `@@toStringTag` it might be better to split these up and make `<feature>.prototype[@@toStringTag]` a separate polyfill for those cases.

Please let me know if you want to help with this or if you need me to run more tests to gather support stats for features.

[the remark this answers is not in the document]
romainmenke: Ideally we also submit patches upstream in MDN for incorrect data :)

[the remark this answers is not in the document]
mhassan1: I've added a `Symbol.toStringTag` check to the detect scripts in `Map`, `Set`, `WeakMap`, and `WeakSet`. I've also updated the configs for `Symbol.toStringTag`, `ArrayBuffer.prototype[@@toStringTag]`, and `TypedArray.prototype[@@toStringTag]` with the versions you added above.

I've also updated the configs for `Map` and `Set` to match the ranges for `@@toStringTag` in MDN. The gaps between support for `prototype[@@iterator]` and `prototype[@@toStringTag]` are not very large, so it seemed better to avoid more polyfills.

I don't see any support data on `WeakMap` or `WeakSet`, so I've left those alone.

polyfills/Map/config.toml
```diff
@@ -24,15 +25,15 @@ notes = [
...
+safari = "<10"
+edge = "<15"
+firefox = "<51"
 chrome = "<49"
 android = "<5.1"
-firefox_mob = "<29"
+firefox_mob = "<51"
 ie_mob = "*"
-opera = "<25"
+opera = "<36"
```
mhassan1, line 35: MDN seems to be wrong here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@toStringTag

I tested this manually in Opera 36, and it is there.

zloirock: [Simple lifehack.](https://github.com/zloirock/core-js/blob/325e059ce48ba81f8ccecb5e0c0e770986cab395/packages/core-js-compat/src/mapping.mjs#L149-L155)

polyfills/ArrayBuffer/prototype/@@toStringTag/config.toml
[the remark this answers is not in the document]
romainmenke: Thank you @mhassan1!
I will check this asap :)

polyfills/Map/config.toml
romainmenke: @zloirock This is awesome!
Thank you 🎉

romainmenke:
Hi @mhassan1,

Sorry that it is taking so long to get this through review and into the library.
Part of this was me getting back up to speed.
My initial review was not entirely accurate but I think I have a better understanding of this change now.

I am hesitant to merge this change as is because it touches a lot of features where the native implementation is significantly better than the polyfill. For example, a polyfill for `Set` or `Map` can never be as performant as a native implementation.

Larger polyfills also impacts the polyfill bundle size which in turn affects parsing and executing size.

With that in mind I prefer having separate polyfills for features like `toStringTag`.
These do not affect the internals of the main polyfill and also do not require access to private values or methods.

Is this something you want to tackle?

I understand that you have already spend significant time on this feature and 
I do want to thank you again for contributing to the library.

JakeChampion:
> For example, a polyfill for `Set` or `Map` can never be as performant as a native implementation.

I think our Map performance has been performant ever since https://github.com/Financial-Times/polyfill-library/pull/293 landed. We haven't yet done the same work for Set though

mhassan1:
I'm happy to create separate polyfills for `toStringTag`. Just a couple of questions:
1. What should we do about `WeakMap`, which already had `toStringTag` inside the main polyfill? I think moving it from the main polyfill would be a breaking change for anyone using `WeakMap` but not `WeakMap.prototype.toStringTag`.
2. What should we do about `*Iterator.prototype.toStringTag` (e.g. `MapIterator`)? Does it make sense for those to also be separate polyfills?

romainmenke:
Good points!
So it is a compromise either way 🤔 

In that case I think we should merge this change as it is now, without splitting it up into more polyfills.

polyfills/TypedArray/prototype/@@toStringTag/config.toml
```diff
@@ -0,0 +1,22 @@
...
+ie = "*"
+ie_mob = "*"
+safari = "<10"
+chrome = "<49"
+firefox = "< 51"
+ios_saf = "< 10"
+opera = "<36"
+samsung_mob = "<5"
+android = "*"
+bb = "10 - *"
```
romainmenke, line 22:
```suggestion
android = "*"
bb = "10 - *"
chrome = "<49"
edge = "<15"
edge_mob = "<15"
firefox = "<51"
ie = "*"
ie_mob = "*"
opera = "<36"
op_mob = "<36"
safari = "<10.0"
ios_saf = "<10.0"
samsung_mob = "<5"
```

polyfills/ArrayBuffer/prototype/@@toStringTag/config.toml
```diff
@@ -0,0 +1,21 @@
...
+ie = "*"
+ie_mob = "*"
+safari = "<10"
+chrome = "<49"
+firefox = "< 51"
+ios_saf = "< 10"
+opera = "<36"
+samsung_mob = "<5"
+android = "*"
+bb = "10 - *"
```
romainmenke, line 21:
```suggestion
android = "*"
bb = "10 - *"
chrome = "<49"
edge = "<15"
edge_mob = "<15"
firefox = "<51"
ie = "*"
ie_mob = "*"
opera = "<36"
op_mob = "<36"
safari = "<10.0"
ios_saf = "<10.0"
samsung_mob = "<5"
```

polyfills/Map/config.toml
```diff
@@ -28,14 +29,14 @@ docs = "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global
...
-firefox = "<36"
-firefox_mob = "<29"
+edge = "<15"
+firefox = "<51"
+firefox_mob = "<51"
 ie = "*"
 ie_mob = "*"
-opera = "<25"
+opera = "<36"
 op_mob = "*"
```
romainmenke, line 38:
```suggestion
opera = "<36"
op_mob = "<36"
```

polyfills/Set/config.toml
```diff
@@ -24,14 +25,14 @@ notes = [
...
-firefox = "<36"
-firefox_mob = "<29"
+edge = "<15"
+firefox = "<51"
+firefox_mob = "<51"
 ie = "*"
 ie_mob = "*"
-opera = "<25"
+opera = "<36"
 op_mob = "*"
```
romainmenke, line 34:
```suggestion
opera = "<36"
op_mob = "<36"
```

polyfills/Symbol/toStringTag/config.toml
```diff
@@ -8,12 +8,12 @@ docs = "https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Ob
...
 bb = "10 - *"
 chrome = "<49"
-edge = "<13"
-edge_mob = "<13"
+edge = "<15"
+edge_mob = "<15"
 firefox = "<51"
 ie = "*"
 ie_mob = "*"
 opera = "<37"
```
romainmenke, line 16:
```suggestion
opera = "<37"
op_mob = "<37"
```

Should this be `<36`?

polyfills/WeakMap/config.toml
```diff
@@ -22,12 +22,12 @@ spec = "https://tc39.github.io/ecma262/#sec-weakmap-objects"
...
+chrome = "<49"
+firefox = "<51"
+firefox_mob = "<51"
 ie = "*"
 ie_mob = ">=10"
-opera = "<25"
-safari = "<9.0"
-ios_saf = "<9.0"
+opera = "<36"
+safari = "<10.0"
```
romainmenke, line 31:
```suggestion
opera = "<36"
op_mob = "<36"
safari = "<10.0"
```

polyfills/WeakSet/config.toml
```diff
@@ -12,20 +12,21 @@ dependencies = [
...
+chrome = "<49"
+firefox = "<51"
+firefox_mob = "<51"
 ie = ">=7"
 ie_mob = ">=10"
-opera = "<25"
-safari = "<9.0"
-ios_saf = "<9.0"
+opera = "<36"
+safari = "<10.0"
```
romainmenke, line 30:
```suggestion
opera = "<36"
op_mob = "<36"
safari = "<10.0"
```

romainmenke (commented):
A few final tweaks and 1 question for `Symbol.toStringTag` which has `Opera <37`

polyfills/Symbol/toStringTag/config.toml
mhassan1: 👍

romainmenke:
running CI : https://github.com/Financial-Times/polyfill-library/actions/runs/2775698723

polyfills/TypedArray/prototype/@@toStringTag/polyfill.js
```diff
@@ -0,0 +1,70 @@
...
+		var name = O._name;
+		// 5. Assert: Type(name) is String.
+		if (Type(name) !== 'string') {
+			throw TypeError();
+		}
+		// 6. Return name.
+		return name;
+	}
+
+	_get.name = 'get [Symbol.toStringTag]'
```
romainmenke, line 25:
```suggestion
	Object.defineProperty(_get, 'name', {
		value: 'get [Symbol.toStringTag]',
		writable: false,
		enumerable: false,
		configurable: true
	});
```

I think this will solve the CI failures

zloirock: It's not always possible. Old engines (IIRC FF <38, Chrome <43, Safari <10) functions `.name` is non-writable non-configurable.

romainmenke: @zloirock Is there a better way to set `.name` or should we wrap `Object.defineProperty` in `try/catch` and remove the failing test?

zloirock: Nope, I don't know a better way. I use try / catch and `defineProperty` and test it only when it's possible to define.

romainmenke: Thank you for this feedback and keeping an eye on all these changes @zloirock :)

@mhassan1 Do you have enough info to get this working?

romainmenke:
CI passed 🎉 
@mhassan1 Is this ready to be merged?
_I assume it is, but want to make sure :)_

mhassan1:
LGTM!

romainmenke (approved):
LGTM!

Thank you @mhassan1 for this change and everyone for the reviews and feedback!

PR #1188, opened 2022-05-19
Fixes `TypedArray.prototype.at` to avoid adding `at` to `Object.prototype`

mhassan1:
The current logic in `TypedArray.prototype.at` adds `at` to the parent prototype of `Int8Array.prototype`; in IE11, this is actually `Object.prototype`, and we don't want to add `at` there. This PR checks that it is not `Object.prototype`; if it is, it adds `at` directly to `Int8Array.prototype`.


polyfills/TypedArray/prototype/at/polyfill.js
```diff
@@ -21,7 +21,9 @@
 		return Get(O, ToString(k));
 	}
 
-	if ('__proto__' in self.Int8Array.prototype) {
+	// IE11, and potentially other browsers, have `Int8Array.prototype` inherit directly from `Object.prototype`
```
chee, line 24:
```suggestion
	// in IE11 `Int8Array.prototype` inherits directly from `Object.prototype`
```

let's not say any more than we know

mhassan1:
@chee Is anything blocking this PR?

romainmenke:
Running CI : https://github.com/Financial-Times/polyfill-library/actions/runs/2650687032

JakeChampion (approved):

romainmenke:
@mhassan1 Thank you for this update!

PR #1200, opened 2022-07-12
fix weekly test suite

romainmenke:
Remnant of https://github.com/Financial-Times/polyfill-library/pull/1199

### DIFF
--- a/.github/workflows/test-polyfills-weekly.yml
+++ b/.github/workflows/test-polyfills-weekly.yml
@@ -11,7 +11,6 @@ concurrency:
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


romainmenke:
@JakeChampion if you have time to review?

JakeChampion:
I think the reason this workflow fails is due to the BrowserStack token being invalid (it would have likely been disabled when I left the FT)

@romainmenke do you know if the token in the GitHub Secrets has been updated?

JakeChampion (approved):

romainmenke:
> do you know if the token in the GitHub Secrets has been updated?

The token was updated by @chee a few weeks ago and still works 🎉 

This job only failed because I missed this one line when I updated the workflow to use native concurrency controls for GitHub workflows.

PR #1201, opened 2022-07-12
fix HTMLElement.prototype.inert

romainmenke:
_tricky issue this one._

The polyfill first added `inert` to `Element.prototype`.
This was incorrect and it should have been `HTMLElement.prototype`.

Part of the issue is how users include this feature in their bundle.
By adding an alias for `Element.prototype.inert` we can make this non breaking.

A more pressing concern is that we are now serving an incorrect polyfill to all browsers with a broken detect. This might cause users to depend on the behaviour of the polyfill instead of the native feature. Users will not know that this is happening.

@JakeChampion thoughts?


JakeChampion (approved):

JakeChampion:
@romainmenke I don't quite understand sorry, what do you mean by a broken feature detect?

romainmenke:
> what do you mean by a broken feature detect?

The current (main) detect.js is :
`'Element' in self && 'inert' in Element.prototype`

But the actual feature shipped as `HTMLElement.prototype.inert`.
`'Element' in self && 'inert' in Element.prototype` is always `false` and the polyfill is always applied even with the gated flag.

JakeChampion:
> But the actual feature shipped as `HTMLElement.prototype.inert`.

oh no. I think we can and should change this to be  `HTMLElement.prototype.inert` and perhaps not include an alias back to `Element.prototype.inert` - that way we won't keep the incorrect polyfill around. For polyfill.io users, they can already select which version of the polyfill-library to use, which means it is less likely that the removal of `Element.prototype.inert` will cause them issues as they would need to explicitly change their polyfill-library version (or have chosen to always use the latest version with the risk of breaking changes).

I'm also happy to keep the alias around and remove it in the next major version release of polyfill-library, whenever we decide to do the next one ☺️

romainmenke:
I've kept the alias for now but added a not that it should be removed.
I've also opened an issue to try and track the things we want to cleanup for a next major. (easy to lose track and forget things)

Thank you for the review 🙇

PR #1204, opened 2022-07-24
validate browser configs

romainmenke:
Analysing and validating the browser configs is too difficult at the moment.
Both for contributors and reviewers the chances of a typo or overlooking something are too great.

Common issues :
- typo's `11.2` vs. `12.1`
- copy/paste mistakes
- browser name keys are undocumented and often misunderstood `op_mob` vs. `op_mini` and `android` vs. `chrome`
- different range styles `<=12` vs. `<13` and `<5` vs. `< 5`
- versions listed in ranges do not exist
- MDN is sometimes incorrect
- non-obvious that you need to map some browsers manually via their engine versions
- browser X doesn't have support today but no one updated it after support landed (`*` everywhere)
- order of browsers in configs

--------

Things that might help :

- lint and format the ranges
- validate browser versions
- suggest values for missing browser entries based on engine mapping
- track and report some stats

More challenging :

It is near impossible to figure out the intent behind certain suspect configs.

`ie = "8 - 11"` vs. `safari = "7 - 12"`

The first is easily understood as any IE 11, even if it has a higher build number (`11.1234`)

The second is often used when MDN says support starts at `12.1`, but that range would actually still match `12.1`. The intent likely was `>=7 <=12.0.0` but this was never written down and these range syntaxes are not obvious or well known.

Part of the issue is that Safari versioning is different from the other browsers.
Other browsers do not have minor versions with new features.

> Note : 
> Samsung seems to have a similar versioning scheme as Safari and we should also switch to `major.minor` to make this more clear

-------

I also fixed the scripts related to `update-polyfill-targeting.js` and made it possible to run these per browser.

In practice it didn't work very well for me.
It takes a very long time to run all the tests and I did not fully trust the output.
Making a tweak to a config and rerunning everything to verify also takes to long to make this a viable flow for contributors.

I might revisit this later and check if it can be useful for maintainers.

-------

Parts :

1) https://github.com/Financial-Times/polyfill-library/pull/1204/commits/397302ee33e3a7a58cd8c37321dfc118989a992d

- android versions stop at 4.4.4, taking this into account changes a lot of android versions to `*`
- ios_chr is legacy and has been removed
- normalise range syntax and whitespace
- fix obvious typos

2) https://github.com/Financial-Times/polyfill-library/pull/1204/commits/a40020d378fcc96e0d6172f4a20475f763c5f6c3

- consistent sorting of browser configs

3) https://github.com/Financial-Times/polyfill-library/pull/1204/commits/bd99cc1d3047a3a20051edfc1080c97383c42f18

always record the minor version for Safari:
- minor versions in Safari are equivalent to major versions in other engines
- explicit notations makes all configs behave more similarly in semver

4) https://github.com/Financial-Times/polyfill-library/pull/1204/commits/ce2a302e675e3cb5fb62e90412b149096ebdf74c

fix edge configs :
- chromium edge is normalised to chrome
- any edge configs with versions higher than 18 have been fixed

5) https://github.com/Financial-Times/polyfill-library/pull/1204/commits/1bf24d051adde244478e847fa01d3487a4a8fe5d

`core-js` static mapping :
- this is hard coded and will need to be updated from time to time
- I've verified each change it suggested and this is very accurate


romainmenke:
This is starting to improve a bit : 

```
{
	unbounded: 822,
	missing: 593,
	withWarnings: 2,
	unknownVersions: 84
}
```

romainmenke:
https://github.com/Financial-Times/polyfill-library/actions/runs/2745221272

Screenshot 2022-07-27 at 10 30 51

.github/contributing.md
```diff
@@ -73,7 +73,6 @@ Browsers that can be listed:
 - edge
 - edge_mob
 - firefox
-- ios_chr
```
JakeChampion, line 76: thank you for removing this 👍

tasks/lint-config/for-each-polyfill-config.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const { promisify } = require('util');
+const glob = promisify(require('glob'));
+const globOptions = { cwd: process.cwd() };
+
+function forEachPolyfillConfigPath(callback) {
+	return glob('polyfills/**/config.toml', globOptions).then((files) => {
+		return Promise.all(files.map((file) => {
+			return callback(file);
```
JakeChampion, line 10: This is really nice 😄

test/polyfills/remotetest.js
```diff
@@ -146,7 +146,7 @@ async function main() {
 		feature = `&feature=${Object.keys(modified.affectedPolyfills).join(',')}`;
 	}
 
-	const includePolyfills = "includePolyfills=" + (mode !== "control" ? "yes" : "no");
+	const includePolyfills = "includePolyfills=" + (mode === "control" ? "no" : "yes");
```
JakeChampion, line 149: that is simpler to understand, thank you ☺️

test/polyfills/compat.js
```diff
@@ -30,6 +36,12 @@ function buildData(support, browserName, version) {
 	};
 }
 
+function trimFeatureName(name) {
+	// When a test times the feature name will be incorrect :
```
JakeChampion, line 40:
```suggestion
	// When a test times out the feature name will be incorrect :
```

test/polyfills/browsers.toml
```diff
@@ -92,27 +93,42 @@ browsers = [
...
+  # "edge/94.0",
+  # "edge/95.0",
+  # "edge/96.0",
+  # "edge/97.0",
+  # "edge/98.0",
+  # "edge/99.0",
+  # "edge/100.0",
+  # "edge/101.0",
+  # "edge/102.0",
+  # "edge/103.0",
```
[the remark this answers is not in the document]
romainmenke: The UA normaliser converts these to Chrome versions.
But maybe better to test a few versions?

- edge/80.0 (earliest in browerstack)
- edge/103.0 (latest at the time of updating `browsers.toml`)

This should catch regressions if the UA normaliser ever fails to map these.

polyfills/AbortController/config.toml
```diff
@@ -14,18 +14,19 @@ repo = "https://github.com/mysticatea/abort-controller"
 docs = "https://developer.mozilla.org/en-US/docs/Web/API/AbortController"
 
 [browsers]
-edge = "12 - 15"
+android = "*"
+chrome = "4 - 65"
```
JakeChampion, line 18: The baseline version for chrome support in polyfill-library is 29
https://github.com/Financial-Times/polyfill-useragent-normaliser/blob/master/data.json#L11

polyfills/AbortController/config.toml
```diff
@@ -14,18 +14,19 @@ repo = "https://github.com/mysticatea/abort-controller"
...
+edge = "<16"
 edge_mob = "*"
-ie = "10 - *"
-ie_mob = "10 - *"
 firefox = "2 - 56"
```
JakeChampion, line 21: The baseline for firefox is 38 
https://github.com/Financial-Times/polyfill-useragent-normaliser/blob/master/data.json#L15

We could rewrite this as <57

polyfills/AggregateError/config.toml
```diff
@@ -7,17 +7,18 @@ license = "MIT"
 docs = "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError"
 
 [browsers]
-android = "<85"
+android = "*"
 bb = "*"
 chrome = "<85"
+edge = "*"
```
JakeChampion, line 13: According to MDN, edge added support in 85 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError#browser_compatibility

test/polyfills/browsers.toml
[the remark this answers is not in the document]
JakeChampion: I like that idea!

polyfills/AggregateError/config.toml
JakeChampion: nvm on this, I understand why now!

>    chromium edge is normalised to chrome
>    any edge configs with versions higher than 18 have been fixed

JakeChampion (approved):

tasks/lint-config/for-each-polyfill-config.js
romainmenke: Working in a context with an older node version does make me appreciate top level `await` just a little bit more each time :D

test/polyfills/compat.js
romainmenke: Good catch! thank you 🙇

polyfills/AbortController/config.toml
romainmenke: I've created a separate issue to tackle these : https://github.com/Financial-Times/polyfill-library/issues/1205

I suspect there will be a lot of these cases.

test/polyfills/browsers.toml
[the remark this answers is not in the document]
romainmenke: I've added those two versions to the list.

romainmenke:
full run with polyfill combinations also passed : https://github.com/Financial-Times/polyfill-library/actions/runs/2748067239

chee:
wow

PR #1209, opened 2022-08-02
Update Safari support for Intl.ListFormat

pyrocat101:
According to the [release note](https://developer.apple.com/documentation/safari-release-notes/safari-14_1-release-notes), Safari >= 14.1 has native support for `Intl.ListFormat`.

### DIFF
--- a/polyfills/Intl/ListFormat/config.toml
+++ b/polyfills/Intl/ListFormat/config.toml
@@ -26,7 +26,7 @@ ie = ">=9"
 ie_mob = ">=9"
 opera = "<60"
 op_mob = "<51"
-safari = "*"
+safari = "<14.1"
 ios_saf = "<14.5"
 samsung_mob = "<11"
 


pyrocat101:
Ok, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat#browser_compatibility), the `Intl.ListFormat` still requires macOS Big Sur or higher. Now I am not sure if this change still makes sense.

romainmenke:
@pyrocat101 Thank you for this pull request.

https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Intl/ListFormat/config.toml#L15

Unfortunately this feature depends on the operating system and mac os Catalina does not support it. For this issue I am hoping that Safari 16 won't be installable on mac os Catalina, but I haven't been able to verify this yet. We will likely have to wait until September/October for the official releases.

If Safari 16 is not installable we will be able to set this version to `<16`. 🤞

pyrocat101:
Makes sense. Closing the PR now :)

Issue #1212, opened 2022-08-04
Polyfill combination issue with `TypedArray` `toLocaleString`


romainmenke:
see : https://github.com/Financial-Times/polyfill-library/runs/7677101952?check_suite_focus=true

```
 - firefox/50.0:
    -> TypedArray.prototype.toLocaleString toLocaleString has the same result as Array.prototype.toLocaleString
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=TypedArray.prototype.toLocaleString
       Missing locale data for 
```

mhassan1:
I can't replicate this error in my own BrowserStack, and I see that this test passed on the [branch](https://github.com/Financial-Times/polyfill-library/runs/7657613540?check_suite_focus=true).

@romainmenke Do you know what `Missing locale data for` means?

romainmenke:
@mhassan1 Thank you for keeping an eye on this :)

It is indeed not obvious why this fails in that specific test run.
All tests in a regular run only include the needed polyfills for the feature being tested.

But users can include other polyfills and polyfills can break each other.
So we have a separate test run that is configured to load all the polyfills together.

This is a very slow process and I usually only run it before a release to prevent regressions.

Adding `polyfillCombinations=yes` to a test url will trigger this mode.

-------

For this specific issue I have found that including one more dependency should fix it : https://github.com/Financial-Times/polyfill-library/pull/1213

I still need to run the full suite on this branch, but I hope that will pass.
Is this the correct fix? (always open to suggestions :) )

mhassan1:
Thanks for the explanation. Sounds good to me.

PR #1213, opened 2022-08-04
TypedArray.prototype.toLocaleString : Intl interop

romainmenke:
add `Intl.NumberFormat.~locale.en` as a dependency

fixes : #1212


romainmenke:
all green : https://github.com/Financial-Times/polyfill-library/actions/runs/2798977699

JakeChampion (approved):
Nice work catching this! Perhaps we should have a full polyfill combination run happen when we make a release branch? I guess it'd require either detecting that package.json.version had changed or we would need to use a special branch name which we configured a GitHub workflow to run on

PR #1216, opened 2022-08-05
Adds polyfill for `Object.prototype.toString`

mhassan1:
This PR adds a polyfill for `Object.prototype.toString` that includes `Symbol.toStringTag` support.


romainmenke:
Test fails with : 

```
Failures:
 - ie/9.0:
    -> Object.prototype.toString returns the class of an object without toStringTag
       http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Object.prototype.toString
       [object Null] == [object Undefined]
```

mhassan1:
@romainmenke Test failure should be resolved. This is ready for review.

JakeChampion (approved):

PR #1218, opened 2022-08-05
Revert "TypedArray.prototype.toLocaleString : Intl interop"

romainmenke:
Reverts Financial-Times/polyfill-library#1213

This seems to have serious performance regression :/


romainmenke:
my bad, wrong PR to revert, this one is fine :)

PR #1186, opened 2022-05-16
Bump @formatjs/intl-datetimeformat

plesiecki:
Hello,
I would like to update the `@formatjs/intl-datetimeformat` dependency as [it polyfills the `Date` class better from version 5.0.2.](https://github.com/formatjs/formatjs/pull/3559)

PR #1219, opened 2022-08-05
Revert "Bump @formatjs/intl-datetimeformat"

romainmenke:
Reverts Financial-Times/polyfill-library#1186

This change seems to have a serious performance regression.
Tests with polyfill combinations always give a timeout now in Chrome, Firefox and IE.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -338,35 +338,13 @@
       }
     },
     "@formatjs/intl-datetimeformat": {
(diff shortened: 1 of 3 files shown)


JakeChampion (approved):
For those following along from home, the upstream ticket is https://github.com/formatjs/formatjs/issues/3747

PR #1220, opened 2022-08-05
validate browser configs (part 2)

romainmenke:
- fill in blanks for `op_mob` and `firefox_mob`
- validate `ios_saf` versions
- use `major.minor` as the significant version for `samsung_mob` (same as `safari`)
- a few configs looked suspicious and have been fixed
- a few features shipped in `safari` in the past few months

more improvements to the lint config script with the final results looking like this :

```md
### Browsers
- 83 missing
- 2 unknown
- 19 unbounded ('*' or '>5')

### Meta
- 27 missing docs link

### Log
- 4 warnings
- 2 info messages
```


test/polyfills/browsers.toml
```diff
@@ -229,9 +229,9 @@ browsers = [
   "firefox/100.0",
   "firefox/101.0",
   "firefox/102.0",
+  "ie/9.0",
```
romainmenke, line 232: In theory it is faster (on average) to start with slow jobs if there are less slow jobs than workers.

_There likely is a good term for this effect._

JakeChampion (approved):

PR #1222, opened 2022-08-10
Adds `x` and `y` to `IntersectionObserverEntry`

mhassan1:
This PR adds `x` and `y` to all instances of `IntersectionObserverEntry` in the `IntersectionObserver` polyfill.

Resolves https://github.com/Financial-Times/polyfill-library/issues/116.

### DIFF
--- a/polyfills/IntersectionObserver/polyfill.js
+++ b/polyfills/IntersectionObserver/polyfill.js
@@ -418,6 +418,8 @@ IntersectionObserver.prototype._getRootRect = function() {
 		var html = document.documentElement;
 		var body = document.body;
 		rootRect = {
+			x: 0,
+			y: 0,
 			top: 0,
 			left: 0,
 			right: html.clientWidth || body.clientWidth,
@@ -449,6 +451,8 @@ IntersectionObserver.prototype._expandRectByRootMargin = function(rect) {
 	};
 	newRect.width = newRect.right - newRect.left;
 	newRect.height = newRect.bottom - newRect.top;
+	newRect.x = newRect.left;
+	newRect.y = newRect.top;
 
 	return newRect;
 };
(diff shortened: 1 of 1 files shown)


romainmenke:
https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly

> [DOMRectReadOnly.top](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly/top) Read only
Returns the top coordinate value of the DOMRect (usually the same as y.)

When is it not the same?

https://drafts.fxtf.org/geometry/#DOMRect

> The y attribute, on getting, it must return the [y coordinate](https://drafts.fxtf.org/geometry/#rectangle-y-coordinate) value. For the [DOMRect](https://drafts.fxtf.org/geometry/#domrect) interface, setting the [y](https://drafts.fxtf.org/geometry/#dom-domrect-y) attribute must set the y coordinate to the new value.

>The top attribute, on getting, must return the [NaN-safe minimum](https://drafts.fxtf.org/geometry/#nan-safe-minimum) of the [y coordinate](https://drafts.fxtf.org/geometry/#rectangle-y-coordinate) and the sum of the y coordinate and the [height dimension](https://drafts.fxtf.org/geometry/#rectangle-height-dimension).

Is it correct that `y` can be `NaN-unsafe` and `top` can not?

mhassan1:
`y` is different from `top` when `height` is negative; for example,
```js
{
  y: 10
  top: 9
  height: -1
}
```
In the case of this polyfill, it doesn't seem like `height` will ever be negative, so they will always be the same.

I'm not sure if any of the `top` values in this polyfill could ever be `NaN`, but if they could, what would we use for `y`, since `y` is derived from `top`?

romainmenke:
I agree that the difference won't matter in the context of this polyfill.
Thank you for helping me understand it better :)

romainmenke (approved):
LGTM! Thank you 🎉

Issue #214, opened 2019-06-03
feature request: nextElementSibling, previousElementSibling on CharacterData


kapouer:
In IE, nextElementSibling is only implemented for Element, not CharacterData.
https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling has a nice summary with a suggested polyfill.

kapouer:
Polyfill for both properties:
```
(function (arr) {
  ['previous', 'next'].forEach(function(name) {
    var keyEl = name + 'ElementSibling';
    var key = name + 'Sibling';
    arr.forEach(function (item) {
      var proto = item.prototype;
      if (proto.hasOwnProperty(keyEl)) {
        return;
      }
      Object.defineProperty(proto, keyEl, {
        configurable: true,
        enumerable: true,
        get: function () {
          var el = this;
          while (el = el[key]) {
            if (el.nodeType === 1) {
              return el;
            }
          }
          return null;
        },
        set: undefined
      });
    });
  });
})([Element, CharacterData]);
```

PR #1224, opened 2022-08-11
Adds `CharacterData.prototype` polyfills

mhassan1:
This PR adds polyfills for `CharacterData.prototype`:
* `after`
* `before`
* `nextElementSibling`
* `previousElementSibling`
* `remove`
* `replaceWith`

This list includes all properties of `CharacterData.prototype` that may be missing from the browsers we target, based on https://developer.mozilla.org/en-US/docs/Web/API/CharacterData#browser_compatibility.

Resolves https://github.com/Financial-Times/polyfill-library/issues/214.

### DIFF
--- a/polyfills/CharacterData/prototype/after/config.toml
+++ b/polyfills/CharacterData/prototype/after/config.toml
@@ -1,0 +1,22 @@
+aliases = []
+dependencies = [ "Element.prototype.after" ]
+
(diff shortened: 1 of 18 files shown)


JakeChampion (approved):

romainmenke:
Thank you @mhassan1 🙇

PR #1225, opened 2022-08-18
improve `Set` performance

romainmenke:
Not a real fix.
This uses `Map` and the `Map` polyfill.

I am curious how much this helps


romainmenke:
_definitely not better_

PR #1227, opened 2022-08-26
Use web-animations-next

KaelWD:
The basic version doesn't include Animation.finished: https://github.com/web-animations/web-animations-js/issues/33

### DIFF
--- a/polyfills/WebAnimations/config.toml
+++ b/polyfills/WebAnimations/config.toml
@@ -26,4 +26,4 @@ samsung_mob = "<5.0"
 
 [install]
 module = "web-animations-js"
-paths = [ "web-animations.min.js" ]
+paths = [ "web-animations-next.min.js" ]


romainmenke:
@KaelWD thank you for your contribution!

To make sure we don't break anything that works today and that the updated polyfill works as intended we would need to do a few more things :

1 Currently there are no tests for this feature here.
That makes it impossible to judge if this change is breaking or not.
We should look at [WPT](https://github.com/web-platform-tests/wpt) and at least a some test coverage for the old polyfill.

2 Change the polyfill to the `next` version

3 Add tests for the newly added sub features

4 Update `config.toml` so that browser support stats and aliasses are up to date.

Is this something you have time for?

KaelWD:
No I don't really have time for that, sorry. 

> The next version is not a more feature rich variant of the same polyfill but rather an updated version and this might be breaking

Where is that quote from? [experimental.md](https://github.com/web-animations/web-animations-js/blob/dev/docs/experimental.md) says "Contains all of web-animations.min.js plus features that are still undergoing discussion or have yet to be implemented natively"

romainmenke:
> Where is that quote from?

It is not a quote, but my reasoning for being cautious here.
The fact that the `next` version never became the main version indicates that it never was safe or stable enough to do so.

Maybe there was no issue and the authors just moved on to other projects but we have no way to determine this without writing a test suite based on WPT and the spec.

Maybe someone else will have time to do this work :)

PR #1235, opened 2022-10-03
Revert "Bump minimist and mocha"

romainmenke:
Reverts Financial-Times/polyfill-library#1233

I don't think we can bump mocha 7 major versions without making any changes :)


JakeChampion (approved):
😅

Issue #1240, opened 2022-11-02
Add HTMLFormElement.requestSubmit Polyfill


andrew-pyle:

JakeChampion:
Howdy, I've not looked at the polyfill myself but thought I'd mention we have contribution guides for adding a new polyfill if you wanted to give it a go 🤞 https://github.com/Financial-Times/polyfill-library/blob/master/.github/contributing.md#adding-a-new-polyfill

mhassan1:
@andrew-pyle I can take a stab at it, unless you wanted to.

andrew-pyle:
@mhassan1 @JakeChampion I actually started working on this about an hour ago. I think I can finish it today.

PR #1241, opened 2022-11-04
Add HTMLFormElement.prototype.requestSubmit Polyfill

andrew-pyle:
Regarding Issue #1240 

First time contributing. I was a little confused on the browser compatibility tables across the Internet, MDN, caniuse, https://github.com/mdn/browser-compat-data. I deleted the browsers which I couldn't find data for. For instance, I deleted `edge_mob` from config.toml. Maintainers, can you look over the versions to see if they look right?


polyfills/HTMLFormElement/prototype/requestSubmit/polyfill.js
```diff
@@ -0,0 +1,68 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE. */
+
+/* global self */
```
romainmenke, line 23:
```suggestion
/* global */
```

romainmenke (commented):
@andrew-pyle Thank you for this 🎉 

The tests include some JS that is too modern for all browsers we run tests in :
- `let` -> `var`
- `for ... of`
- `() => ` -> `function()`

Everything should work in IE9 without requiring any extra build steps :)

andrew-pyle:
I see. I suppose that makes sense. The tests can't run on old browsers if the tests have code old browsers can't run.

OK I'll need some more time to remove those features. I used quite a few multiline backtick strings, and I'm out of time at work today. I will probably be able to update this code later tonight (America/Chicago timezone).

polyfills/HTMLFormElement/prototype/requestSubmit/config.toml
```diff
@@ -0,0 +1,20 @@
...
+firefox = "<75"
+firefox_mob = "<79"
+ie = "*"
+ie_mob = "*"
+opera = "<63"
+op_mob = "<64"
+op_mini = "*"
+safari = "<16"
+ios_saf = "<16"
+samsung_mob = "<12"
```
romainmenke, line 20:
```suggestion
[browsers]
android = "*"
chrome = "<76"
edge = "*"
edge_mob = "*"
firefox = "<75"
firefox_mob = "<79"
ie = "*"
ie_mob = "*"
opera = "<63"
op_mob = "<64"
op_mini = "*"
safari = "<16.0"
ios_saf = "<16.0"
samsung_mob = "<12.0"
```

- we record `edge < 79` as `edge *` (same for `android`)
- `safari`, `ios_saf`, and `samsung_mob` use `major.minor` notation
- `edge_mob` can have the same value as `edge`

romainmenke:
Hi @andrew-pyle I tried to run the tests but everything failed.
Do you have a way to run the tests locally in an older browser like IE?

If not I will try to do so asap and get you some meaningful debug info.

andrew-pyle:
@romainmenke I don't have a way to test locally on IE, at least that I know. I am working on an M1 iMac. If you know a way to get IE installed on macOS, I'd be glad to do it. Oh—I also have a 32-bit Debian server accessible as well. Can we install IE on Debian?

romainmenke:
We use https://www.browserstack.com which makes it relatively simple to debug things locally if you have an account.

I will try to run the tests on my end and get you that debug info :)

romainmenke:
Hi @andrew-pyle,

I've looked into the test failures and unfortunately it is not one obvious and simple thing.

Issue one :
- the `iframe` elements in the tests trigger a load of `http://bs-local.com:9876/test`
- `http://bs-local.com:9876/test` includes all tests (including this one)
- 🔥 we now have an infinite loop, not a particularly hot one, but still something that browsers will eventually kill.

Issue two :
Seeing a `object doesn't support this` error message.
Because of the issue above I haven't been able to pinpoint the cause.
For some reason I suspect `DOMException`, maybe older IE doesn't support `new DOMException()`? I might be totally off here, so maybe best to ignore my hunch.

Is it possible to rewrite the tests so that we are sure there can not be an infinite loop through iframes?

andrew-pyle:
@romainmenke Thanks for the debugging information!

I'm acting as more of an editor than an author for this polyfill, so I don't know how the tests are designed, precisely. I only adapted the tests from the Web Platform Tests (WPT) for `HTMLFormElement.prototype.requestSubmit()`. I'll be glad to look into the test structure, but it will be several days until I can. Maybe Friday or the weekend.

--------------------------

**Some ideas from me:**

- Issue 1: The `<iframe>` is loaded only one test, like this. I'm not sure how a `/test` gets involved. There is a `<form>` in that test which has an action  of `<form action="/common/blank.html"></form>`. Maybe that is related?
  ```js
  document.body.insertAdjacentHTML(
    "afterbegin",
    '<iframe name="iframe" src="about:blank"></iframe>'
  );
  ```
  Link to test with `<iframe>`: https://github.com/andrew-pyle/polyfill-library/blob/master/polyfills/HTMLFormElement/prototype/requestSubmit/tests.js#L230 
- Issue 2: Several possibilities:
  - **I use `new FormData()` per the original WPT. [I see now this is unsupported in IE < 11](https://caniuse.com/mdn-api_formdata), so `FormData` should be removed!**
  - I do access a`DomException.name` property in several tests, to detect a "NotFoundError", per the original WPT.
  - I also use `new URLSearchParams()`, which I think is supported in IE 9

--------------------------

If you want to look at original source of the tests, here is a link to the test running live https://wpt.live/html/semantics/forms/the-form-element/form-requestsubmit.html.

The source code for the `HTMLFormElement.prototype.requestSubmit()` WPT is here: https://github.com/web-platform-tests/wpt/blob/master/html/semantics/forms/the-form-element/form-requestsubmit.html

andrew-pyle:
@romainmenke I've made some headway on the issues with the tests for this PR. I've replaced several things with IE9-friendly counterparts:
- `UrlSearchParams.get()` with `RegExp.test()`
- Made the tests abpit FormData behavior skip if the browser does not have FormData (`!('FormData' in window)`). Is this a good solution, actually?

Unfortunately, I've not solved the infinite loop problem. I do understand it better, now. The problem occurs when the test submits a form to the `<iframe>` on the page. This is necessary, so that the test can examine the search params in the URL generated by the form submission, like`?s=v`. The`<form>`'s action attribute is something that should be a 404 in the testing context (`/common/blank.html`), so I am not sure why the <iframe> gets sent to the test-runner page: `/test`. @romainmenke Do you know of anything about Browserstack which would redirect a GET for a 404 to the test page itself? That's my only guess.

At any rate, to finish these WPT tests, we will need to find a way to evaluate the `x-www-form-urlencoded` search params generated after the polyfill submits a form.

andrew-pyle:
I am revisiting this after being stumped, giving up, and leaving it alone for a while.

I'd really like to see `requestSubmit()` on Polyfill.io. Not having `requestSubmit()` in the polyfill list prevents me from relying on Polyfill.io at work. We need to polyfill `requestSubmit()`.

I am quite confident that the polyfill code itself is accurate. It has [~30,000 weekly downloads on NPM](https://www.npmjs.com/package/form-request-submit-polyfill). But the Pull Request tests aren't passing.

I can't seem to figure out how to set up the tests to run in the test environment. I was able to run the tests locally, in all the browsers which are installed on my Mac computer. I didn't understand how to set up a Browserstack account and run the tests on old/obscure browsers there.

Can the maintainers provide some assistance here? Once again, I am motivated to get this merged, but I don't know how to proceed.

romainmenke:
This is still broken on older browsers like IE 11.
I will try to make time for this in one of the next weekends.

JakeChampion:
@romainmenke I plan to update polyfill.io in a month or so time (once I'm the full owner) to the latest and greatest polyfill-library

polyfills/HTMLFormElement/prototype/requestSubmit/tests.js
```diff
@@ -1,28 +1,60 @@
...
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
romainmenke, line 34: Wrapped to 80 chars, no content changes

polyfills/HTMLFormElement/prototype/requestSubmit/tests.js
```diff
@@ -1,28 +1,60 @@
...
+			var input = document.createElement("input");
+			input.setAttribute("form", "the-form");
+
+			document.body.appendChild(form);
+			document.body.appendChild(input);
+
+			supportsFormAssociatedElements = input.form === form;
+			document.body.removeChild(form);
+			document.body.removeChild(input);
+		})();
```
romainmenke, line 56: Older browsers do not support associating elements to a `<form>` with the `form` attribute.

polyfills/HTMLFormElement/prototype/requestSubmit/tests.js
```diff
@@ -38,15 +70,35 @@ describe("HTMLFormElement.prototype.requestSubmit", function () {
...
 	it("Passing an element which is not a submit button should throw", function () {
 		document.getElementById("test-container").insertAdjacentHTML(
 			"afterbegin",
-			'<form><input type="reset"><input type="text"><button type="reset"></button><button type="button"></button></form>'
+			'<form action="/empty-document.html" target="iframe">' +
+			'  <input type="reset">' +
+			'  <input type="text">' +
+			'  <button type="reset"></button>' +
+			'  <button type="button"></button>' +
+			'</form>'
```
romainmenke, line 94: I formatted a lot of these because it was hard to read these one liners.

polyfills/HTMLFormElement/prototype/requestSubmit/tests.js
```diff
@@ -59,158 +111,145 @@ describe("HTMLFormElement.prototype.requestSubmit", function () {
...
-		invalidControl.addEventListener("invalid", function () {
+		control1.addEventListener("invalid", function () {
 			didDispatchInvalid = true;
 		});
 
-		form.requestSubmit();
-		proclaim.isTrue(didDispatchInvalid);
-	});
-
-	it("requestSubmit() doesn't run form submission reentrantly", function () {
```
romainmenke: The `doesn't do X reentrantly` tests can never pass.
Only a native implementation can keep track of the entire `submission` process.

It requires knowing when a `submit` process starts and ends.
Any calls to `requestSubmit` during such a process should be skipped, but we can't implement this.

polyfills/HTMLFormElement/prototype/requestSubmit/tests.js
```diff
@@ -227,16 +266,15 @@ describe("HTMLFormElement.prototype.requestSubmit", function () {
 	it("The value of the submitter should be appended, and form* attributes of the submitter should be handled.", function (done) {
 		document.getElementById("test-container").insertAdjacentHTML(
 			"afterbegin",
-			'<form action="/common/blank.html"><input required><input type=submit formnovalidate formtarget=iframe name=s value=v></form>'
+			'<form action="/empty-document.html" target="iframe"><input required><input type=submit formnovalidate name="prop" value="e277b63f-a0b1-437e-a557-51c854f7d19b"></form>'
```
romainmenke, line 269: `<form action="/empty-document.html" target="iframe">`

All forms need to have this action and target.
Without this a submission will trigger a navigation and the test server will respond with a page that includes all tests for all features, including this one. (e.g. an infinite loop)

test/polyfills/server.js
```diff
@@ -61,6 +61,9 @@ const cacheFor1Day = cache("1 day", () => true, {
 
 app.get(["/test"], createEndpoint(runnerTemplate));
 app.get(["/iframe.html"], createEndpoint(testIframeTemplate));
+app.get(["/empty-document.html"], cacheFor1Day, (request, response) => {
+	response.sendFile(path.resolve(__dirname, "./empty-document.html"));
+});
```
romainmenke, line 66: Empty document response to prevent cycles.

JakeChampion (approved):
Everything passed on CI for this ☺️

test/polyfills/server.js
JakeChampion: Nice work spotting and fixing this @romainmenke 🙌

romainmenke:
@andrew-pyle Thank you for you contribution!

andrew-pyle:
Thank you @romainmenke for rescuing this PR from the infinite loop! I didn't know several of those things, about `<form>`, reentrancy, and the test environment node server.

PR #1242, opened 2022-11-07
Adds structuredClone. Fixes #1147

### LINKED ISSUE #1147: Feature/Polyfill request: structuredClone
As the title says, I would like to have a [structuredClone](https://github.com/ungap/structured-clone/) polyfill. At the time of writing, all browsers have implemented this API in their nightly releases, Firefox has shipped it to stable in Firefox 94. Additionally, Node 17 and Deno 1.14 have implemented this API.

CheloXL:
This adds a basic implementation of structuredClone. Of course not all features can be polyfilled, but this one works quite well.

### DIFF
--- a/polyfills/structuredClone/config.toml
+++ b/polyfills/structuredClone/config.toml
@@ -4,6 +4,9 @@ spec = "https://html.spec.whatwg.org/multipage/structured-data.html#dom-structur
 docs = "https://developer.mozilla.org/en-US/docs/Web/API/structuredClone"
 license = "MIT"
 repo = "https://github.com/ungap/structured-clone/"
+notes = [
+	"This polyfill currently does not support cloning: Blob, File, FileList, ImageBitmap, ImageData, and ArrayBuffer. TypedArrays are supported without major issues, but int and uint 8/16/32 are the only safely supported TypedArrays."
+]
 
 [browsers]
 android = "<107"
--- a/polyfills/structuredClone/config.toml
+++ b/polyfills/structuredClone/config.toml
@@ -1,8 +1,9 @@
 dependencies = [
 	"Map",
 	"Set",
+	"Object.keys",
 	"String.prototype.includes",
-	"Object.keys"
+	"RegExp.prototype.flags"
 ]
 spec = "https://html.spec.whatwg.org/multipage/structured-data.html#dom-structuredclone"
 docs = "https://developer.mozilla.org/en-US/docs/Web/API/structuredClone"
--- a/polyfills/structuredClone/config.toml
+++ b/polyfills/structuredClone/config.toml
@@ -1,5 +1,7 @@
-aliases = []
-dependencies = []
+dependencies = [
+	"Map",
+	"Set"
+]
 spec = "https://html.spec.whatwg.org/multipage/structured-data.html#dom-structuredclone"
 docs = "https://developer.mozilla.org/en-US/docs/Web/API/structuredClone"
 license = "MIT"
--- a/polyfills/structuredClone/config.toml
+++ b/polyfills/structuredClone/config.toml
@@ -1,0 +1,23 @@
+aliases = []
+dependencies = []
+spec = "https://html.spec.whatwg.org/multipage/structured-data.html#dom-structuredclone"
+docs = "https://developer.mozilla.org/en-US/docs/Web/API/structuredClone"
+license = "MIT"
+repo = "https://github.com/ungap/structured-clone/"
+
+[browsers]
+android = "<107"
+bb = "*"
+chrome = "<98"
+edge = "<98"
+edge_mob = "<98"
+firefox = "<94"
+firefox_mob = "<106"
+ie = "*"
+ie_mob = "*"
+opera = "<84"
+op_mob = "*"
+op_mini = "*"
+safari = "<15.4"
+ios_saf = "<15.4"
+samsung_mob = "<18"
--- a/polyfills/structuredClone/config.toml
+++ b/polyfills/structuredClone/config.toml
@@ -1,6 +1,8 @@
 dependencies = [
 	"Map",
-	"Set"
+	"Set",
+	"String.prototype.includes",
+	"Object.keys"
 ]
 spec = "https://html.spec.whatwg.org/multipage/structured-data.html#dom-structuredclone"
 docs = "https://developer.mozilla.org/en-US/docs/Web/API/structuredClone"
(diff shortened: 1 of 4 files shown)


ScottAwesome:
This looks like it largely lifts the implementation of structured clone from [ungap](https://www.npmjs.com/package/@ungap/structured-clone)

Therefore, I imagine it has the following limitations, correct?

>not supported yet: Blob, File, FileList, ImageBitmap, ImageData, and ArrayBuffer, but typed arrays are supported without major issues, but u/int8, u/int16, and u/int32 are the only safely supported

This should be noted in some way

CheloXL:
> This looks like it largely lifts the implementation of structured clone from [ungap](https://www.npmjs.com/package/@ungap/structured-clone)

Indeed. I first tried to use it directly from npm, but it's coded in es6 and build failed as the build pipeline doesn't do any traspiling and failed when trying to minify it.

> Therefore, I imagine it has the following limitations, correct?
> 
> > not supported yet: Blob, File, FileList, ImageBitmap, ImageData, and ArrayBuffer, but typed arrays are supported without major issues, but u/int8, u/int16, and u/int32 are the only safely supported
> 
> This should be noted in some way

Correct. I don't know where to put that.

polyfills/structuredClone/config.toml
```diff
@@ -0,0 +1,23 @@
+aliases = []
+dependencies = []
+spec = "https://html.spec.whatwg.org/multipage/structured-data.html#dom-structuredclone"
+docs = "https://developer.mozilla.org/en-US/docs/Web/API/structuredClone"
+license = "MIT"
+repo = "https://github.com/ungap/structured-clone/"
+
```
JakeChampion, line 7:
```suggestion
notes = [
  "This polyfill currently does not support cloning: Blob, File, FileList, ImageBitmap, ImageData, and ArrayBuffer. TypedArrays are supported without major issues, but int and uint 8/16/32 are the only safely supported TypedArrays."
]
```

CheloXL: Done

CheloXL:
Mmmm.. not sure about the failing linting... What should I do with the BigInt object? I can't put it on dependecies, since there is no BigInt polyfill. Also not sure about the other linting issues.
obj["return"] is written that way to not have issues with IE and keywords. Not sure what to do about Unsafe usage of ThrowStatement.

polyfills/structuredClone/polyfill.js
```diff
@@ -1,438 +1,271 @@
...
+				case ARRAY: {
+					var arr = as([], index);
+					for (i = 0, l = value.length; i < l; ++i)
+						arr.push(unpair(value[i]));
+					return arr;
+				}
+
+				case OBJECT: {
+					var object = as({}, index);
+					var keys = Object.keys(value);
```
JakeChampion, line 45: `Object.keys` should be listed in the `dependencies` array in `polyfills/structuredClone/config.toml`

polyfills/structuredClone/polyfill.js
```diff
@@ -1,438 +1,271 @@
...
+						set.add(unpair(value[i]));
+					return set;
+				}
+
+				case ERROR: {
+					return as(new env[value.name](value.message), index);
+				}
+
+				case BIGINT:
+					return as(BigInt(value), index);
```
JakeChampion, line 83: Note: I don't think we have a `BigInt` polyfill

polyfills/structuredClone/polyfill.js
```diff
@@ -1,438 +1,271 @@
...
+				return [REGEXP, EMPTY];
+
+			case "Map":
+				return [MAP, EMPTY];
+
+			case "Set":
+				return [SET, EMPTY];
+		}
+
+		if (asString.includes("Array"))
```
JakeChampion, line 129: `String.prototype.includes` should be listed in the `dependencies` array in `polyfills/structuredClone/config.toml`

polyfills/structuredClone/tests.js
```diff
@@ -11,9 +10,9 @@ describe('structuredClone', function () {
 		proclaim.arity(structuredClone, 2);
 	});
 
-	const date = new Date;
+	var date = new Date();
 
-	const obj = {
+	var obj = {
 	  arr: [],
 	  bigint: 1n,
```
JakeChampion, line 17: This won't work as the syntax is not supported in all the browsers we run our tests within. I think we'd need to use the function instead of the literal. I.E. `BigInt(1)` instead of `1n`.

polyfills/structuredClone/tests.js
```diff
@@ -1,5 +1,4 @@
-/* eslint-env mocha, browser */
-// eslint-disable-next-line no-unused-vars
+/* eslint-env es2020, mocha, browser */
```
JakeChampion, line 1: We don't want new syntax in the tests, could you please remove `es2020` from the env declaration - our tests need to run in es5 browsers which don't support the newer syntax

JakeChampion (dismissed):

polyfills/structuredClone/polyfill.js
CheloXL: It's not possible to polyfill BigInt. Anyways, if you run this polyfill on IE11, that path should never be hit, as there is no way you have a BigInt in an in-memory object.

CheloXL:
Fixed one of the errors in tests about a missing `flags` property on Regex. Some tests fail with errors in browserstack configuration.
I'm not sure about the failing one "null == null" on really old browsers. I don't know what that means... null should be null but it is not? I'm guessing that's on the line that has `proclaim.equal(deserialized.null, null);`.

I'm aware of `null == undefined` and `null !== undefined`, but that's not what I see in the test. Does anyone has a clue of what could be failing?

JakeChampion:
> Fixed one of the errors in tests about a missing `flags` property on Regex. Some tests fail with errors in browserstack configuration. I'm not sure about the failing one "null == null" on really old browsers. I don't know what that means... null should be null but it is not? I'm guessing that's on the line that has `proclaim.equal(deserialized.null, null);`.
> 
> I'm aware of `null == undefined` and `null !== undefined`, but that's not what I see in the test. Does anyone has a clue of what could be failing?

I'm not quite sure of the reason for the error, but just to be safe I've used square bracket notation on all the properties which are also es3 keywords

CheloXL:
> I'm not quite sure of the reason for the error, but just to be safe I've used square bracket notation on all the properties which are also es3 keywords

Now all tests passed, but linting does not (doesn't like the bracket syntaxt).. I think that should be accepted anyways, as is the only way for that to work on older browsers... right?

CheloXL:
Not sure if I have to do something else... failing check is [update-check-fork](https://github.com/Financial-Times/polyfill-library/actions/runs/3732243442/jobs/6331436640#logs)

JakeChampion (approved):

PR #1257, opened 2023-01-11
fix `Error.cause`

romainmenke:
see : https://github.com/Financial-Times/polyfill-library/actions/runs/3869344912/jobs/6595380297

`_InstallErrorCause` is stateful and keeps track of various prototypes of original `Error` subclasses and `Error` itself.

```js
	var _nativeErrors = {};
	var _errorConstructors = {};
```

This stateful nature also makes the order of the polyfills important.
When applied in the incorrect order `AggregateError` will be polyfilled first and `Error` second.

This breaks the prototype chain in the current setup.

----

This change makes `Error.cause` an explicit dependency of `AggregateError` which ensures that all polyfills are applied in a known order.

There might be a better fix, but `prototype` fiddling has never been my area of expertise :)
Happy to fix this in an other way if someone has a suggestion.

@mhassan1 would be great if you could also take a look?

-----

I will need to run the full test suite to be sure all issues are resolved.
I will post an update here when that has been done.


polyfills/_InstallErrorCause/polyfill.js
```diff
@@ -52,7 +52,7 @@ var _GetErrorConstructorWithCauseInstalled;
 	var _nativeErrors = {};
 	var _errorConstructors = {};
 
-	_GetErrorConstructorWithCauseInstalled = function (name) {
+	_InstallErrorCause = function (name) {
```
mhassan1, line 55: i think this name is confusing, since it's not installing anything

polyfills/AggregateError/config.toml
```diff
@@ -4,6 +4,7 @@ dependencies = [
 	"_ESAbstract.CreateMethodProperty",
 	"_ESAbstract.IterableToList",
 	"_InstallErrorCause",
+	"Error.cause",
```
mhassan1, line 7: i was hoping to avoid this because it's not technically correct. we could simplify things if we want to go this route, but i think we should explore another route, first.

mhassan1:
Are there replication steps for this? I can't replicate it in BrowserStack on Chrome 84.

I don't know if putting `Error.cause` as a dependency is correct; I jumped through some hoops in the original implementation to avoid that.

romainmenke:
> Are there replication steps for this? I can't replicate it in BrowserStack on Chrome 84.

Yes, you can test this with `polyfillCombinations=yes` :
http://bs-local.com:9876/?includePolyfills=yes&always=no&polyfillCombinations=yes

This will cause the test server to bundle all polyfills, not just the requested polyfill and its dependencies.

polyfills/_InstallErrorCause/polyfill.js
romainmenke: Maybe just `_ErrorCause` ?
This would mirror `_DOMTokenList` and `DOMTokenList`

polyfills/AggregateError/config.toml
mhassan1: this change doesn't seem to resolve the problem for me on Chrome 84: http://localhost:9876/test?includePolyfills=yes&always=no&feature=AggregateError&polyfillCombinations=yes

romainmenke: 🤔 seems fine here : https://github.com/Financial-Times/polyfill-library/actions/runs/3895684914/jobs/6651304275

But there are still other issues.

mhassan1:
@romainmenke Here is an alternative fix: https://github.com/Financial-Times/polyfill-library/pull/1258

Issue #1273, opened 2023-01-31
CI to test community submissions is broken


romainmenke:
https://github.com/Financial-Times/polyfill-library/pull/1264

@notlee The CI flows to test community submissions seems to be broken since the recent change.
Can you take a look or do you have a reference/docs for the change?

_failing run : https://github.com/Financial-Times/polyfill-library/actions/runs/4058518902/jobs/6985539450_

notlee:
@romainmenke Sorry about that, this was a dumb change. Let me revert – the risks are low.

Do you know if Github's in-built feature to [approve workflows runs from public forks](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) approves a PR and any future commits, or requires a second approval? I'm guessing now at that's why we have ok-to-test

romainmenke:
It might have changed since `ok-to-test` was added.
I will investigate this.

Would be nice to drop that dependency and rely on native features.

JakeChampion:
ok-to-test is used because the trigger comes from a user who has write permissions which then means the action runs have access to our Secretes -- the built-in approve workflow runs from public forks does not give access to Secrets (for security reasons)

PR #1274, opened 2023-02-01
Revert "backstage: only checkout approved commits when running tests"

notlee:
Reverts Financial-Times/polyfill-library#1264
Closes https://github.com/Financial-Times/polyfill-library/issues/1273


romainmenke (approved):
LGTM!
Thank you @notlee

PR #1275, opened 2023-02-02
map polyfill : ensure that map keys are sufficiently unique

romainmenke:
Before this change map keys for non string types would be the same as string values that happened to be the same as `x.toString()`.

- `99` and `'99'`
- `true` and `'true'`
- `null` and `'null'`
- ...

We now prepend the type name and a unique marker (a newly generated UUID).
This makes the chance of collisions really small.

### DIFF
--- a/polyfills/Map/polyfill.js
+++ b/polyfills/Map/polyfill.js
@@ -31,8 +31,11 @@
 			// Return previously defined hashed key
 			return recordKey[_metaKey];
 		}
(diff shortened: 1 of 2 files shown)


romainmenke:
I am unsure if browsersstack changed something or if they have an outage.
Was running the full test suite and it fails both for this feature and on the main branch.

Going to retry later.

PR #1285, opened 2023-04-17
Adds `change-array-by-copy` polyfills

mhassan1:
This PR adds the following ES2023 polyfills:
* `Array.prototype.toReversed`
* `Array.prototype.toSorted`
* `Array.prototype.toSpliced`
* `Array.prototype.with`
* `%TypedArray%.prototype.toReversed`
* `%TypedArray%.prototype.toSorted`
* `%TypedArray%.prototype.with`


mhassan1:
> https://github.com/Financial-Times/polyfill-library/actions/runs/4762911502

There are a bunch of failures:
* `F must be a constructor`: The `ArrayBuffer` polyfill does not expose `int8Array.constructor.name`
* `With == with`: Missing dependency on `Function.prototype.name`
* `A.sort is not a function`: The `ArrayBuffer` polyfill includes `TypedArray.prototype.sort`, but we don't deliver `ArrayBuffer` to browsers that have `TypedArray` but don't have `TypedArray.prototype.sort` (e.g. [Chrome 44](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/sort#browser_compatibility))
* `[TypedArray].prototype.sort: argument is not a Function object`: We should not pass `undefined` to `sort`

All of these have been fixed except `A.sort is not a function`, which may require some discussion. I think we have a few options:
1. Deliver the `ArrayBuffer` polyfill to versions that do not have `TypedArray.prototype.sort`
2. Write a separate `TypedArray.prototype.sort` polyfill
3. Implement `TypedArray.prototype.sort` inside `TypedArray.prototype.toSorted`

JakeChampion:
> > [Financial-Times/polyfill-library/actions/runs/4762911502](https://github.com/Financial-Times/polyfill-library/actions/runs/4762911502)
> 
> There are a bunch of failures:
> 
>     * `F must be a constructor`: The `ArrayBuffer` polyfill does not expose `int8Array.constructor.name`
> 
>     * `With == with`: Missing dependency on `Function.prototype.name`
> 
>     * `A.sort is not a function`: The `ArrayBuffer` polyfill includes `TypedArray.prototype.sort`, but we don't deliver `ArrayBuffer` to browsers that have `TypedArray` but don't have `TypedArray.prototype.sort` (e.g. [Chrome 44](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/sort#browser_compatibility))
> 
>     * `[TypedArray].prototype.sort: argument is not a Function object`: We should not pass `undefined` to `sort`
> 
> 
> All of these have been fixed except `A.sort is not a function`, which may require some discussion. I think we have a few options:
> 
>     1. Deliver the `ArrayBuffer` polyfill to versions that do not have `TypedArray.prototype.sort`
> 
>     2. Write a separate `TypedArray.prototype.sort` polyfill
> 
>     3. Implement `TypedArray.prototype.sort` inside `TypedArray.prototype.toSorted`

Thanks for the detailed write-up! I'm in favour of option 2. It has precedent as we took that option for Promise.prototype.finally as well -- https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Promise/prototype/finally/config.toml#L9

JakeChampion:
Tests running at <https://github.com/Financial-Times/polyfill-library/actions/runs/4872382344>

romainmenke:
CI all green 🎉

JakeChampion (approved):

PR #1287, opened 2023-05-01
node 16

romainmenke:
see : https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

At some point we will have to switch to a new node version for CI.

In this attempt I've updated various dependencies naively.
I am hoping to update as many dependencies as possible to also get rid of some npm audit nags that only affect dev env's.

-----

Scripts verified :

- [x] lint
- [x] lint-config
- [x] clean
- [x] build
- [x] watch
- [x] fmt
- [x] test-library
- [x] test-polyfills
- [ ] test-all-polyfills (broken) [^2]
- [x] test-end-to-end
- [x] test-node
- [x] test-node-unit
- [x] ~~test-node-coverage~~ (removed) [^1]
- [x] ~~test-coverage~~ (removed) [^1]
- [ ] test [^2]
- [x] prepublishOnly
- [x] create-new-polyfill
- [x] update-browserstack-list
- [x] version

[^1]: these tests were never called. I integrated coverage checking into the normal test commands
[^2]: this is also broken on main, but these commands are not really used in CI or by me.


tasks/buildsources/build.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const path = require('path');
-const makeDirectory = require('mkdirp');
+const makeDirectory = require('mkdirp').mkdirp;
```
JakeChampion, line 4: Node 16 has this as a built-in option on `fs.mkdir(path,{recursive: true})`

romainmenke: Good catch!
I've updated these and a few other operation that work fine with just `fs`.
I didn't touch anything under `lib` because I don't want this change to have any user facing side-effects.

JakeChampion (approved):

PR #1288, opened 2023-05-08
`Array.prototype.sort` : extra tests

romainmenke:
Extra tests for `Array.prototype.sort` to ensure that it is really doing a stable sort.

_Initially created to debug a test failure in `Array.prototype.toSorted`_

-----

Original message :

The stable sort polyfill is causing issues for `toSorted`.
Somehow it gives unexpected results with sparse arrays.

I haven't found the issue yet, but these are some extra tests.

@mhassan1 I've been testing on Edge 15 (but 18 should fail aswel) with this url :
http://bs-local.com:9876/test?includePolyfills=yes&always=no&feature=Array.prototype.toSorted&polyfillCombinations=yes

Maybe you spot something I am overlooking :)

### DIFF
--- a/polyfills/Array/prototype/sort/tests.js
+++ b/polyfills/Array/prototype/sort/tests.js
@@ -89,12 +89,60 @@ it("sorts arrays with comparefn that returns non-number results", function () {
 	);
 });
 
-it("has a stable sort", function () {
-	var obj = {length:3, 0:2, 1:1,2:3};
+it("has a stable sort with array-like objects", function () {
+	var obj = { length: 3, 0: 2, 1: 1, 2: 3 };
 	proclaim.deepStrictEqual(
 		Array.prototype.sort.call(obj, function (a, b) {
 			return a - b;
 		}),
 		obj
 	);
+
+	proclaim.equal(obj[0], 1);
+	proclaim.equal(obj[1], 2);
+	proclaim.equal(obj[2], 3);
+});
+
+it("has a stable sort with arrays that contain duplicate values", function () {
+	var array = [
+		{ unique: 'a', sortValue: 1 },
+		{ unique: 'b', sortValue: 2 },
+		{ unique: 'c', sortValue: 0 },
+		{ unique: 'd', sortValue: 4 },
+		{ unique: 'e', sortValue: 2 },
+		{ unique: 'f', sortValue: 2 }
+	];
+
+	array.sort(function (a, b) {
+		return a.sortValue - b.sortValue;
+	});
+
+	proclaim.equal(array[0].unique, 'c');
+	proclaim.equal(array[1].unique, 'a');
+	proclaim.equal(array[2].unique, 'b');
+	proclaim.equal(array[3].unique, 'e');
+	proclaim.equal(array[4].unique, 'f');
+	proclaim.equal(array[5].unique, 'd');
+});
+
+it("has a stable sort with arrays that contain duplicate values and comparefn occasionally returns NaN", function () {
+	var array = [
+		{ unique: 'a', sortValue: 1 },
+		{ unique: 'b', sortValue: 2 },
+		{ unique: 'c', sortValue: 0 },
+		{ unique: 'd', sortValue: 4 },
+		{ unique: 'e', sortValue: undefined },
+		{ unique: 'f', sortValue: 2 }
+	];
+
+	array.sort(function (a, b) {
+		return a.sortValue - b.sortValue;
+	});
+
+	proclaim.equal(array[0].unique, 'c');
+	proclaim.equal(array[1].unique, 'a');
+	proclaim.equal(array[2].unique, 'b');
+	proclaim.equal(array[3].unique, 'd');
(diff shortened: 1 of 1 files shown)


mhassan1:
This seems to be related to lack of [stable sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility), which is missing in old Edge and Chrome <70. You can see this by running `[6, 4, 5, , 3].sort(function(){return NaN})` on a blank page in latest Chrome (`[ 6, 4, 5, , 3 ]`) vs. Chrome 69 (`[ 6, 4, 5, 3, ]`.

It looks like our `Array.prototype.sort` polyfill doesn't implement stable sort properly, since `[6, 4, 5, , 3].sort(function(){return NaN})` results in `[6, 4, 5, 3, 3]` instead of the expected `[ 6, 4, 5, , 3 ]`.

@JakeChampion It looks like you implemented the `Array.prototype.sort` polyfill. Any thoughts?

@romainmenke If we won't want to wait for `Array.prototypt.sort` polyfill to be fixed, we could simplify the `toSorted` tests to not be dependent on whether stable sort is available or not, since that's beyond the responsibility of the `toSorted` tests.

polyfills/Array/prototype/sort/polyfill.js
```diff
@@ -41,6 +43,9 @@ CreateMethodProperty(Array.prototype, "sort", function sort(compareFn) {
 		});
 		// update the original object (`this`) with the new position for the items
 		// which were moved.
+		this.length = 0;
+		this.length = len;
```
romainmenke, line 47: `this` needs to be reset.
Otherwise sparse arrays will cause items at the end to remain.

`[1,,3]` -> `[1,3,3]`

By resetting we get `[1,3,]`

polyfills/Array/prototype/toSorted/polyfill.js
```diff
@@ -19,7 +19,9 @@ CreateMethodProperty(Array.prototype, 'toSorted', function toSorted(comparefn) {
 	// 8. Repeat, while j < len,
 	while (j < len) {
 		// a. Perform ! CreateDataPropertyOrThrow(A, ! ToString(𝔽(j)), sortedList[j]).
-		CreateDataPropertyOrThrow(A, ToString(j), O[j]);
+		if (Object.prototype.hasOwnProperty.call(this, ToString(j))) {
+			CreateDataPropertyOrThrow(A, ToString(j), O[j]);
+		}
```
romainmenke, line 24: `['a',]`

If we iterate and assign `A[1]` with the value found at index `1` the array doesn't have holes anymore. It will have an assignment of `undefined` at index `1`.

romainmenke:
@mhassan1 Thank you so much!

> will exclude empty slots.

This helped me find out what was going on.
I was trying to trace the values and I was overlooking the empty slots 🙇 

Can you take a look at these proposed changes?

mhassan1:
@romainmenke I just raised an alternate [PR](https://github.com/Financial-Times/polyfill-library/pull/1289), which touches the polyfill more and adds more tests. I am not sure if these two PRs are equivalent...

polyfills/Array/prototype/sort/polyfill.js
```diff
@@ -41,6 +43,9 @@ CreateMethodProperty(Array.prototype, "sort", function sort(compareFn) {
 		});
 		// update the original object (`this`) with the new position for the items
 		// which were moved.
+		this.length = 0;
```
mhassan1, line 46: is it necessary to set the length to `0` first?

polyfills/Array/prototype/toSorted/polyfill.js
mhassan1: let's add a test for this condition. i see modern chrome doing `[6,4,5,,3].toSorted() // -> [3, 4, 5, 6, undefined]`, so i think the current behavior is correct.

polyfills/Array/prototype/sort/tests.js
mhassan1, line 1: i think we are still missing a test for a sparse array

polyfills/Array/prototype/sort/polyfill.js
romainmenke: Unsure 🤔.

I think your changes to `.sort` look better than my changes here.
I will take a fresh look at these tomorrow :)

Thank you for jumping on this so quickly 🎉

romainmenke:
Thank you for all the changes @mhassan1 

Is there anything from this PR that would be useful or has it all been superseded by your fixes?

----

Running the full test suite here : https://github.com/Financial-Times/polyfill-library/actions/runs/4924977370

Browserstack is gain having some issues, but I am fairly confident that everything is fixed now.

mhassan1:
@romainmenke The additional `Array.prototype.sort` tests might be nice; otherwise, no.

romainmenke:
@mhassan1 I've reset this branch and added just the tests, including one more with a sparse array. I think this is good to go.

mhassan1 (approved):

PR #1289, opened 2023-05-08
Fixes `Array.prototype.sort` for sparse arrays

mhassan1:
This PR fixes `Array.prototype.sort` so it properly handles sparse arrays, arrays containing `undefined`, and `comparefn` that returns invalid results. This implementation is based on https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/array-sort.js.

### DIFF
--- a/polyfills/Array/prototype/sort/polyfill.js
+++ b/polyfills/Array/prototype/sort/polyfill.js
@@ -31,22 +31,32 @@ CreateMethodProperty(Array.prototype, "sort", function sort(compareFn) {
 		// if compareFn exists, sort the array, breaking sorting ties by using the
 		// items' original index position.
 
+		var index;
+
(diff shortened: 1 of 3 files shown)


polyfills/Array/prototype/sort/polyfill.js
```diff
@@ -31,22 +31,32 @@ CreateMethodProperty(Array.prototype, "sort", function sort(compareFn) {
...
-				}
+		index = 0;
+		while (index < that.length) {
+			if (that[index]) {
+				this[index] = that[index].item;
 			}
+			index++;
+		}
+		while (index < this.length) {
+			delete this[index++];
```
JakeChampion, line 59: good catch!

JakeChampion (approved):

PR #1293, opened 2023-05-18
docs(readme): added section of AWS Lambda

RodrigoTomeES:


JakeChampion (approved):

README.md
```diff
@@ -84,6 +84,34 @@ Create a polyfill bundle.
...
+```bash
+yarn add -D make-dir fs-extra
+```
+
+```js
+import { copySync } from 'fs-extra/esm';
+import makeDir from 'make-dir';
+
+const DIR_POLYFILLS = './node_modules/polyfill-library/polyfills/__dist';
+const DIR_SERVERLESS = 'LAMBDA_DIR/polyfills/__dist';
```
JakeChampion, line 102: @RodrigoTomeES is `LAMBDA_DIR` meant to be a variable? If not, could we add a comment to explain why it is like this please?

RodrigoTomeES: Hi, LAMBDA_DIR is not a variable, is only an example, I will explain it better

RodrigoTomeES: @JakeChampion how do you see it now?

JakeChampion (approved):

PR #1306, opened 2023-07-31
Working polyfill for console.exception instead of noop function

juliyvchirkov:
Provides working polyfill for ``console.exception`` instead of noop function

The thing is according to **[MDN Web APIs reference]( https://developer.mozilla.org/en-US/docs/Web/API/console)** ``console.exception`` is an obsolete alias for ``console.error`` method

So, if ``error`` exists in ``console``, we can provide it under ``console.exception`` name and thus get this polyfill working. If there's no ``error`` in ``console``, we're falling back to noop ``function exception() {}``

### DIFF
--- a/polyfills/console/exception/polyfill.js
+++ b/polyfills/console/exception/polyfill.js
@@ -1,5 +1,5 @@
 console.exception = function exception() {
-  if ("error" in console) {
-    console.error.apply(this, arguments);
-  }
+    if ("error" in console) {
+        Function.prototype.apply.call(console.error, null, arguments);
+    }
 };
--- a/polyfills/console/exception/polyfill.js
+++ b/polyfills/console/exception/polyfill.js
@@ -1,5 +1,5 @@
 console.exception = function exception() {
     if ("error" in console) {
-        Function.prototype.apply.call(console.error, null, arguments);
+        Function.prototype.apply.call(console.error, console, arguments);
     }
 };
--- a/polyfills/console/exception/polyfill.js
+++ b/polyfills/console/exception/polyfill.js
@@ -1,1 +1,1 @@
-console.exception = function exception() {};
+console.exception = 'error' in console ? console.error : function exception() {};
--- a/polyfills/console/exception/polyfill.js
+++ b/polyfills/console/exception/polyfill.js
@@ -2,4 +2,4 @@ console.exception = function exception() {
   if ("error" in console) {
     console.error.apply(this, arguments);
   }
-);
+};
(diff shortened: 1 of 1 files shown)


polyfills/console/exception/polyfill.js
```diff
@@ -1 +1 @@
-console.exception = function exception() {};
+console.exception = 'error' in console ? console.error : function exception() {};
```
JakeChampion, line 1: I believe rhe function name would be different with this change, it'd be `error` instead of `exception`. 

If we conditionally cal `error` then we can have the function name always be `exception`.
```suggestion
console.exception = function exception() {
  if ("error" in console) {
    console.error.apply(this, arguments);
  }
);
```

juliyvchirkov: @JakeChampion Agree

Or it could be even ``console.error.apply(console, arguments)``

My reason why I had not used this way was the functuion name anyway turns into something like ``console.exception = function n() {}`` after minification, so there's no big difference if the name will be ``error`` or ``n``

zloirock: Note that in some old engines, `console` methods are not `Function` instances -> haven't methods like `.apply`.

juliyvchirkov: @zloirock Thanks a lot for this critical info

I'll rearrange the code now with ``Function.prototype.call.apply(console.error, arguments)``

juliyvchirkov: @zloirock Thanks once again for the remark. Now the polyfill code shouldn't throw even on these rare engines https://github.com/JakeChampion/polyfill-library/pull/1306/commits/bd721604690d51670aed8c7c9e896668df35fdc4

polyfills/console/exception/polyfill.js
```diff
@@ -1,5 +1,5 @@
 console.exception = function exception() {
-  if ("error" in console) {
-    console.error.apply(this, arguments);
-  }
+    if ("error" in console) {
+        Function.prototype.apply.call(console.error, null, arguments);
```
zloirock, line 3: Use `console` instead of `null` - in some old engines it's required.

zloirock: https://github.com/whatwg/console/issues/3

juliyvchirkov: @zloirock lol, [I've mentioned this](https://github.com/JakeChampion/polyfill-library/pull/1306#discussion_r1279948180) a couple of hours ago

juliyvchirkov: @zloirock [Commited](https://github.com/JakeChampion/polyfill-library/pull/1306/commits/a4d5e8826061cc1db419e714786458a9bb33b863), thanks again for noticing

PR #1311, opened 2023-08-16
update browserstack list

JakeChampion:
It seems browserstack have removed some devices/browsers that we were testing against - this pull-requests update our list to test against so that our browserstack integration keeps working


test/polyfills/browsers.toml
```diff
@@ -107,11 +105,19 @@ browsers = [
...
+  "chrome/113.0",
+  "chrome/114.0",
+  "chrome/115.0",
+  "chrome/116.0 beta",
   "edge/15.0",
   "edge/16.0",
   "edge/17.0",
   "edge/18.0",
-  "edge/80.0", # oldest chromium edge version we can test
+  "edge/80.0", "edge/80.0", # oldest chromium edge version we can test
```
romainmenke, line 120:
```suggestion
  "edge/80.0", # oldest chromium edge version we can test
```

test/polyfills/browsers.toml
```diff
@@ -107,11 +105,19 @@ browsers = [
...
   "chrome/107.0",
   "chrome/108.0",
+  "chrome/109.0",
+  "chrome/110.0",
+  "chrome/111.0",
+  "chrome/112.0",
+  "chrome/113.0",
+  "chrome/114.0",
+  "chrome/115.0",
+  "chrome/116.0 beta",
```
romainmenke, line 115: needed?
I think we avoided beta versions previously.
I don't have a strong opinion on this.

PR #1312, opened 2023-08-16
update uglifyjs to v3 which supports more minification optimisations

JakeChampion:


polyfills/esm/polyfill.js
romainmenke, line 1: I think this file was included by accident?

JakeChampion: 🙈

romainmenke (approved):
LGTM!

Nice that this is now using a recent release of uglify-js :)

PR #1320, opened 2023-12-09
update the browsers list

romainmenke:
see : https://github.com/JakeChampion/polyfill-library/pull/1319


tasks/updatebrowserstacklist.js
```diff
@@ -8,23 +8,39 @@ dotenv.config({
 	path: path.join(__dirname, "../../.env")
 });
 const browserStackCredentials = {
-	username: process.env.BROWSER_STACK_USERNAME,
-	password: process.env.BROWSER_STACK_ACCESS_KEY
+	username: process.env.BROWSERSTACK_USERNAME,
+	password: process.env.BROWSERSTACK_ACCESS_KEY
```
romainmenke, line 12: All other scripts use `BROWSERSTACK`, only this script used `BROWSER_STACK`.

tasks/updatebrowserstacklist.js
```diff
@@ -8,23 +8,39 @@ dotenv.config({
...
+	if (aBrowser !== bBrowser) {
+		return aBrowser.localeCompare(bBrowser);
+	}
+
+	if (aVersion !== bVersion) {
+		return semver.lt(semver.coerce(aVersion), semver.coerce(bVersion)) ? -1 : 1;
+	}
+
+	return 0;
+}
```
romainmenke, line 46: This makes it easier to update the list.
It sorts by browser name and then by semver.

This previously did a simpler string sort.

JakeChampion (commented):
Thank you!

JakeChampion (approved):

